PR #6, opened 2017-04-02
ENHANCEMENT Display list of campaigns the item is already in

phalkunz:


src/AddToCampaignHandler.php
```diff
@@ -171,31 +185,70 @@ protected function getObject($id, $class)
...
         );
 
+        // If there's no campaigns to add to, we shouldn't have the dropdown
+        // and add button
+        if ($changeSets->count() === 0) {
+            $campaignAdminUrl = CampaignAdmin::singleton()->config()->get('url_segment');
+            $link = Controller::join_links('admin', $campaignAdminUrl);
+            $noCampaingsText = _t(
+                'Campaigns.AddToCampaignNoCampaignsToAdd',
+                '<p class="lead text-xs-center">This item has been added to all available campaigns.<br/><a href="#" class="add-to-campaign-modal__nav-link">Create a new campaign?</a></p>',
```
flamerohr: fallback doesn't use the `campaignSectionLink` link provided

src/AddToCampaignHandler.php
```diff
@@ -117,6 +119,18 @@ protected function getAvailableChangeSets()
...
     /**
+     * Get changesets that a given object is already in
+     *
+     * @param DataObject
+     * @return ArrayList[ChangeSet]
+     */
+    protected function getInChangeSets($object)
+    {
+        $inChangeSetIDs = array_unique(ChangeSetItem::get_for_object($object)->column('ChangeSetID'));
+        return $this->getAvailableChangeSets()->filter(['ID' => $inChangeSetIDs, 'State' => ChangeSet::STATE_OPEN]);
```
tractorcow: Need to check for $inChangeSetIDs not being empty before doing filter.

src/AddToCampaignHandler.php
```diff
@@ -171,31 +185,70 @@ protected function getObject($id, $class)
...
+        $inChangeSets = $this->getInChangeSets($object);
+        $changeSets = $this->getAvailableChangeSets();
 
-        $campaignDropdown = DropdownField::create('Campaign', '', $changeSets);
+        // Remove campaigns that the $object is already in
+        foreach ($inChangeSets as $cs) {
+            $changeSets->remove($changeSets->find('ID', $cs->ID));
+        }
+
+        $inChangeSetsText = "";
```
tractorcow: How about a "This item is not currently in any campaigns"?

src/AddToCampaignHandler.php
```diff
@@ -171,31 +185,70 @@ protected function getObject($id, $class)
...
         );
 
+        // If there's no campaigns to add to, we shouldn't have the dropdown
+        // and add button
+        if ($changeSets->count() === 0) {
+            $campaignAdminUrl = CampaignAdmin::singleton()->config()->get('url_segment');
+            $link = Controller::join_links('admin', $campaignAdminUrl);
+            $noCampaingsText = _t(
+                'Campaigns.AddToCampaignNoCampaignsToAdd',
+                '<p class="lead text-xs-center">This item has been added to all available campaigns.<br/><a href="{campaignSectionLink}" class="add-to-campaign-modal__nav-link">Create a new campaign?</a></p>',
```
tractorcow: This is only true if this item is in a campaign. If there are no campaigns at all this will still appear, incorrectly.

src/AddToCampaignHandler.php
```diff
@@ -245,7 +298,7 @@ public function addToCampaign($object, $campaignID)
         $request = $this->controller->getRequest();
         $message = _t(
             'AddToCampaign.Success',
-            'Successfully added {ObjectTitle} to {CampaignTitle}',
+            'Successfully added <strong>{ObjectTitle}</strong> to <strong>{CampaignTitle}</strong>',
```
tractorcow: If you want to send back html:
 - set content-type to "text/html" below
 - Ensure you use convert::raw2xml() on all titles

src/AddToCampaignHandler.php
flamerohr: We don't really want to show any message when it's not in any campaigns ;)

src/AddToCampaignHandler.php
phalkunz: Good catch 👍

src/AddToCampaignHandler.php
phalkunz: Done!

src/AddToCampaignHandler.php
phalkunz: Yike! Will fix it.

PR #8, opened 2017-04-06
WIP Resolve list item breaking

clarkepaul:
Small fix to reduce the layout breaking for campaign list items. 

Doc thumbnails in the campaign list require either a CSS class or change to use css background-image as per Files area.

If someone can add a type CSS class I could continue with it, otherwise I pass this on for someone else to finish.

Fixes: https://github.com/silverstripe/silverstripe-framework/issues/6430


clarkepaul:
@phalkunz if you have time can you add a CSS class name to the thumbnail to signify what file type is being viewed so I can style appropriately.

tractorcow:
Already fixed.

PR #17, opened 2017-05-05
Pulls/4.0/5464 duplicate campaign name

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/5464

### DIFF
--- a/src/CampaignAdmin.php
+++ b/src/CampaignAdmin.php
@@ -570,6 +570,7 @@ class CampaignAdmin extends LeftAndMain implements PermissionProvider
     public function save($data, $form)
     {
         $request = $this->getRequest();
+        $errors = null;
 
         // Existing or new record?
         $id = empty($data['ID']) ? 0 : $data['ID'];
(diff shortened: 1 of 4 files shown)


src/CampaignAdmin.php
```diff
@@ -588,12 +589,26 @@ public function save($data, $form)
...
         }
 
-        // save form data into record
-        $form->saveInto($record, true);
-        $record->write();
-        $this->extend('onAfterSave', $record);
-
-        $message = _t('LeftAndMain.SAVEDUP', 'Saved.');
+        $hasExistingName = Changeset::get()
+            ->filter('Name:case', $data['Name'])
```
flamerohr: I would opt for case-insensitivity when filtering for a duplicate.
Having an entry `My case` and allowing for `My Case` to also be created is going to be confusing.

Was this discussed to have it case-sensitive?

phalkunz: No, it wasn't in any discussion. I thought it was a good idea but in hindsight it is rather confusing. Good spotting it :)

PR #21, opened 2017-05-24
Show link icon and total number of linked from and linked to

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6685


client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -66,13 +66,28 @@ class CampaignAdminList extends SilverStripeComponent {
...
 
+  getSelectedItem() {
+    const itemId = this.props.campaign.changeSetItemId;
+    const items = this.getItems() || [];
+    let selected = null;
+
+    selected = items.find(item => itemId === item.ID);
+
+    if (!selected && items.length > 0) {
+      selected = items[0];
```
flamerohr: was it agreed to have the first item as the selected by default?

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -66,13 +66,28 @@ class CampaignAdminList extends SilverStripeComponent {
     this.props.breadcrumbsActions.setBreadcrumbs(breadcrumbs);
   }
 
+  getSelectedItem() {
+    const itemId = this.props.campaign.changeSetItemId;
+    const items = this.getItems() || [];
+    let selected = null;
+
+    selected = items.find(item => itemId === item.ID);
```
flamerohr: check if itemId is null/falsey before trying to find it?

client/src/containers/CampaignAdmin/tests/CampaignAdminItem-test.js
```diff
@@ -0,0 +1,149 @@
...
+    it('should show link icon when the item is being from the selected item', () => {
+      props.isLinked = true;
+      cmp = ReactTestUtils.renderIntoDocument(
+        <CampaignAdminItem {...props} />
+      );
+      tooltip = ReactTestUtils
+        .scryRenderedDOMComponentsWithClass(cmp, 'campaign-admin__item-links');
+
+      expect(tooltip).toHaveLength(1);
+      expect(tooltip[0]).toBeInstanceOf(HTMLElement);
```
flamerohr: would it be possible check the content or something, e.g. `tooltip[0].Text`?

src/CampaignAdmin.php
```diff
@@ -250,6 +250,7 @@ protected function getChangeSetItemResource(ChangeSetItem $changeSetItem)
         $hal = [
             '_links' => [
```
flamerohr: I would like to investigate removing the underscore prefix in this pull request, it'd be good to fix this before getting merged

client/dist/js/vendor.js
```diff
@@ -1 +1,99 @@
...
\ No newline at end of file
+<<<<<<< HEAD
```
phalkunz: I just tried re-`build`, it didn't seem to generate the `vendor.js` file. Do we still need this file?

client/src/containers/CampaignAdmin/CampaignAdminItem.js
```diff
@@ -1,12 +1,66 @@
...
+
+    return numReferTo || 0;
+  }
+
+  /**
+   * @return integer
+   */
+  getNumReferredBy() {
+    const numReferredBy = (
+      this.props.item._links &&
```
phalkunz: I didn't create `_links` but I can rename it

client/src/containers/CampaignAdmin/CampaignAdminItem.js
```diff
@@ -1,12 +1,66 @@
...
+
+  /**
+   * @return string
+   */
+  getReferToTooltipText() {
+    const numReferTo = this.getNumReferTo();
+
+    const text = numReferTo < 2 ?
+      i18n._t('CampaignAdmin.LINKED_TO_SINGULAR', 'Links to %s item') :
+      i18n._t('CampaignAdmin.LINKED_TO_PLURAL', 'Links to %s items');
```
phalkunz: Is there a story for pluralisation already?

client/src/containers/CampaignAdmin/CampaignAdminItem.js
```diff
@@ -40,11 +94,49 @@ class CampaignAdminItem extends SilverStripeComponent {
...
+    if ((this.props.selected && numReferTo + numReferredBy > 0) || this.props.isLinked) {
+      const linksClasses = [
+        'list-group-item__info',
+        'campaign-admin__item-links',
+        this.props.isLinked ?
+          'campaign-admin__item-links--is-linked' :
+          'campaign-admin__item-links--has-links',
+      ];
+
+      links = (
```
phalkunz: Great! will do :)

src/CampaignAdmin.php
phalkunz: Will do

client/src/containers/CampaignAdmin/CampaignAdminList.js
phalkunz: I think that would be unnecessary. What would be an edge case for not having the check?

client/src/containers/CampaignAdmin/CampaignAdminList.js
phalkunz: Based on the current implementation (before this change). If you can load a campaign, the first item in the campaign is selected.

client/src/containers/CampaignAdmin/tests/CampaignAdminItem-test.js
phalkunz: Actually the name is wrong and misleading – `tooltip` should be `link` or `links` instead.

client/dist/js/vendor.js
flamerohr: ah, it could be that it's no longer needed and `vendor.js` can be deleted (due to the webpack2 upgrade)

client/src/containers/CampaignAdmin/CampaignAdminItem.js
flamerohr: You're right, probably isn't a story for the javascript side (there's one for the php side which was done), I'll be happy to raise one :)
but yeah, I don't like the hardcoding for pluralisation, it only caters for english really.

client/src/containers/CampaignAdmin/CampaignAdminList.js
flamerohr: Oh ok, we can leave this then :D

client/src/containers/CampaignAdmin/tests/CampaignAdminItem-test.js
flamerohr: That's cool, could be renamed, but I was more looking for a more robust unit test.
So rather than checking for a generic `HTMLElement` which can contain anything, look for a bit more specifically the first item title that's supposed to be rendered.

src/CampaignAdmin.php
flamerohr: thanks!

client/src/containers/CampaignAdmin/tests/CampaignAdminItem-test.js
phalkunz: Actually, this line is just redundant to the one above it. I'll just remove it. The purpose of this test case is to make sure the link is element is there. The tooltip text and and number are tested in the cases below.

client/src/containers/CampaignAdmin/CampaignAdminItem.js
tractorcow: 👍 for not fixing pluralisation to work in english only. Many locales have more than 2 plural forms, some only 1.

src/CampaignAdmin.php
```diff
@@ -287,14 +288,24 @@ protected function getChangeSetItemResource(ChangeSetItem $changeSetItem)
...
         }
 
+        $referToItems = $changeSetItem->findReferenced();
+        $referTo = [];
+        foreach ($referToItems as $referToItem) {
+            $referTo[] = [
+                'ID' => $referToItem->ID
+            ];
+        }
+        $hal['links']['refer_to'] = $referTo;
```
tractorcow: the inverse of `referenced_by` is `references` not `refer_to`.

src/CampaignAdmin.php
```diff
@@ -287,14 +288,24 @@ protected function getChangeSetItemResource(ChangeSetItem $changeSetItem)
...
-                $hal['_links']['referenced_by'] = $referencedBy;
+                $hal['links']['referenced_by'] = $referencedBy;
             }
         }
 
+        $referToItems = $changeSetItem->findReferenced();
+        $referTo = [];
+        foreach ($referToItems as $referToItem) {
+            $referTo[] = [
+                'ID' => $referToItem->ID
```
tractorcow: You cannot identify objects only by ID; Different classes can have the same ID. E.g. SiteTree.ID = 1 and File.ID = 1

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -114,14 +137,25 @@ class CampaignAdminList extends SilverStripeComponent {
           itemClassNames.push('active');
         }
 
+        let isLinked = !!selectedItemsLinkedTo.find(
+            linkToObj => linkToObj.ID === parseInt(item.ObjectID, 10));
+
+        isLinked = isLinked || selectedItemsLinkedFrom.find(
+            linkFromObj => linkFromObj.ObjectID === item.ObjectID);
```
tractorcow: See my earlier comment; You are matching on ID here only, and not on class.

src/CampaignAdmin.php
tractorcow: Inverse of `referenced_by` is `references`, not `refer_to`

src/CampaignAdmin.php
tractorcow: See example code from ChangeSet.php

```php
$referenced[$key] = [
    'ObjectID' => $referee->ID,
    'ObjectClass' => $referee->baseClass(),
];
```

src/CampaignAdmin.php
```diff
@@ -193,7 +193,7 @@ protected function getListResource()
     protected function getChangeSetResource(ChangeSet $changeSet)
     {
         $hal = [
-            '_links' => [
```
tractorcow: This might make sense for react.js , but it breaks HAL which this the format this code is written for.

http://stateless.co/hal_specification.html

package.json
```diff
@@ -60,7 +61,8 @@
       "client/src"
     ],
     "modulePaths": [
-      "client/src"
+      "client/src",
+      "../silverstripe-admin/client/src"
```
tractorcow: Don't assume silverstripe-admin is one directory up. Also include `silverstripe-admin/client/src` for installation of admin as a nested module.

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -114,14 +137,25 @@ class CampaignAdminList extends SilverStripeComponent {
           itemClassNames.push('active');
         }
 
+        let isLinked = !!selectedItemsLinkedTo.find(
+            linkToObj => linkToObj.ID === parseInt(item.ObjectID, 10));
+
+        isLinked = isLinked || selectedItemsLinkedFrom.find(
```
tractorcow: Why do you only show selectedItemsLinkedFrom if selectedItemsLinkedTo is empty?

As an analogy, this would be only showing child pages in the site tree for pages that have no parent.

client/src/containers/CampaignAdmin/tests/CampaignAdminItem-test.js
```diff
@@ -1,8 +1,6 @@
 /* global jest, describe, beforeEach, it, expect */
 
-jest.unmock('react');
-jest.unmock('react-addons-test-utils');
-jest.unmock('../CampaignAdminItem');
```
phalkunz: Thanks :)

client/src/containers/CampaignAdmin/CampaignAdminList.js
flamerohr: The check means that it does not need to loop through the whole list to find an empty item

src/CampaignAdmin.php
```diff
@@ -287,14 +288,24 @@ protected function getChangeSetItemResource(ChangeSetItem $changeSetItem)
...
+                    'ObjectID' => $referencedItem->ObjectID
                 ];
             }
             if ($referencedBy) {
-                $hal['_links']['referenced_by'] = $referencedBy;
+                $hal['links']['referenced_by'] = $referencedBy;
             }
         }
 
+        $referToItems = $changeSetItem->findReferenced();
```
tractorcow: There's already a `->References()` relation you can use that has a pre-generated list of changesetitems. That's the other side of ReferencedBy().

src/CampaignAdmin.php
flamerohr: Hmm yeah ok...
The code is getting quite confusing switching between a reference link, the ChangeSetItemID and ObjectID...
@phalkunz are we able to standardise this to use either ChangeSetItemID or ObjectID+ObjectClass, and not use a mix of these?
Sorry for the mix up, it was quite confusing to read through the code review once @tractorcow pointed out the HAL specs...

Perhaps ChangeSetItemID would be the best, because it's always a unique identifier.

tractorcow:
@phalkunz my suggestion is to change ->findReferenced() (list of objects) to -> References() (list of changesets).

Then after that, the ID of each item will be the changesetID, not the ID of the object in the changeset item. This will be a correct identifier, as you have a fixed class (ChangeSetItem), and will be the same type as the ReferencedBy() relation you iterate over above, so your code can be consistent.

src/CampaignAdmin.php
phalkunz: @tractorcow That does not work. `findReferened()` get all implicit **objects** for this change while `References()` is a list of implicit **items** required by this change.

client/src/containers/CampaignAdmin/CampaignAdminItem.js
phalkunz: I've reverted the removal of `_` due HAL specs

client/src/containers/CampaignAdmin/CampaignAdminItem.js
phalkunz: Done!

src/CampaignAdmin.php
```diff
@@ -287,14 +288,24 @@ protected function getChangeSetItemResource(ChangeSetItem $changeSetItem)
...
             if ($referencedBy) {
                 $hal['_links']['referenced_by'] = $referencedBy;
             }
         }
 
+        $referToItems = $changeSetItem->References();
+        $referTo = [];
+        foreach ($referToItems as $referToItem) {
+            $referTo[] = [
+                'ChangeSetItemID' => $referToItem->ID,
```
tractorcow: Probably should have the `href` in here as well, but it's not critical.

tractorcow:
Code looks fine. @flamerohr please test locally and confirm it works as expected. :D

Issue #16, opened 2017-05-04
On successful submission of create campaign form, there's no success message


phalkunz:
# Original issue
The reason is user is redirected to the edit form so the message set on create form never gets a chance to show it to user.

@tractorcow said:
> Store the message in redux and show it then
> you'll need a custom state / message handler for campaign admin
> otherwise, you'll need a way of making the create-form set the success message on the edit form
> ...
> Easiest solution is to make the submit handler for create set a session message on the edit form 
> but i say that's the easiest, because it can be done in php :smile:
> best way is to do it in JS and not rely on session

Related: https://github.com/silverstripe/silverstripe-framework/issues/5464

# Acceptance criteria
* Once a campaign has been successfully created the user is directed to the (empty) campaign view and reassured the campaign has been created with a success message on the newly created campaign
* The user has a clear indication of what to do next (eg. adding content to the campaign), and how to add items from the newly created campaign
* The campaign is not publishable until there is content in the campaign
* When a user navigates to a campaign they are presented with guides on how to use the campaign
* The user has the option to select an item from the campaign (if items exist) so they can preview them
* When campaign settings are edited and then saved there is a clear indication that they have been saved

# Tasks
* [x] User flow provided to guide development
* [x] The "Add campaign" button is updated to have the label "Add new", as per pages pattern
* [x] The "Save" button on the create campaign form is updated to have a "Create" label
* [x] Show empty sections within the campaign for Pages and Files, and provide a prompt to add content
* [x] Add visual (to be provided) to the initial view of a campaign—regardless of whether it is a new campaign or not
* [x] Designer to provide the visual to go in place of the preview panel
* [x] On the campaign settings edit screen the save button (when clicked) label updates to "Saved" with a tick icon
* [x] If the campaign settings are edited we detect the changes and swap the action label (and colour) from "Saved" to "Save as per page actions when saved/published
* [x] The "Publish campaign" button is disabled if there is nothing in the campaign

# Designs
Interactive mockup to outline the new flow https://invis.io/S4CMLL6J2#/243581682_Add_Campaign_6

Design Asset:
![image](https://user-images.githubusercontent.com/8085295/28350182-3daa4f96-6c9b-11e7-8b0c-6690a2af59e0.png)

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-campaign-admin/pull/37
- [x] https://github.com/silverstripe/silverstripe-admin/pull/167
[...]

PR #37, opened 2017-07-21
FIX Create campaign success message

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-campaign-admin/issues/16

Requires https://github.com/silverstripe/silverstripe-admin/pull/178 to be merged first

Merge these after this
- https://github.com/silverstripe/silverstripe-asset-admin/pull/573
- https://github.com/silverstripe/silverstripe-cms/pull/1918


client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -352,6 +436,7 @@ function mapStateToProps(state, ownProps) {
     record: record || {},
     campaign: state.campaign,
     treeClass,
+    newItemCreated: state.campaign.newItemCreated,
```
flamerohr: proptypes definition for this

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -239,7 +303,14 @@ class CampaignAdminList extends SilverStripeComponent {
 
     let actionProps = {};
 
-    if (this.props.record.State === 'open') {
+    if (!items || items.length === 0) {
+      actionProps = Object.assign(actionProps, {
+        title: i18n._t('CampaignAdmin.PUBLISHCAMPAIGN'),
```
flamerohr: add translation default fallback

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -196,18 +226,57 @@ class CampaignAdminList extends SilverStripeComponent {
...
+          <p>Loading...</p>
+        </div>
+      );
+    } else if (!this.getItems() || this.getItems().length === 0) {
+      preview = (
+        <div
+          className="flexbox-area-grow
+          fill-height
+          preview
+          campaign-admin__campaign-preview
```
flamerohr: shouldn't have new lines in double quoted strings...
move these to a variable or classnames object

client/src/state/campaign/CampaignReducer.js
```diff
@@ -6,6 +6,7 @@ const initialState = deepFreeze({
   changeSetItemId: null,
   isPublishing: false,
   view: null,
+  newItemCreated: false,
```
flamerohr: the name for this is quite confusing, since this is a boolean type, it should describe the question like `isNewItem` or `hasNewItem`.
`newItemCreated` implies it's an object that was just created.

client/src/state/campaign/CampaignActions.js
```diff
@@ -65,3 +65,16 @@ export function publishCampaign(publishApi, recordType, campaignId) {
...
   };
 }
+
+/**
+ * Set new campaign created flag
+ *
+ * @param {boolean}
+ * @return {Object}
+ */
+export function setNewCampaignCreated(newItemCreated) {
```
flamerohr: naming is inconsistent with what it's doing, maybe `setIsNewItem` or `markHasCreatedItem`

having both `New` and `Created` is double up on meaning.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -119,7 +120,8 @@ class CampaignAdmin extends SilverStripeComponent {
           // open the new campaign in edit mode after save completes
           const sectionUrl = this.props.sectionConfig.url;
           const id = response.record.id;
-          this.props.router.push(`${sectionUrl}/set/${id}/edit`);
+          this.props.campaignActions.setNewCampaignCreated(true);
```
flamerohr: Since the `id` is available, it would be more extensible to include the record or even just the id in the record, rather than just a boolean.
In this case, you could ignore the `is` or `has` prefix I mentioned below. But try to keep the naming consistent still `setNewItem` is perfectly fine.

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -196,18 +226,51 @@ class CampaignAdminList extends SilverStripeComponent {
...
+        <div className={previewClasses}>
+          <p>Loading...</p>
+        </div>
+      );
+    } else if (!this.getItems() || this.getItems().length === 0) {
+      preview = (
+        <div className={previewClasses}>
+          <h2 className="campaign-admin__empty-heading">Getting started</h2>
+          <p className="campaign-admin__empty-info">
+            Select <strong>Add to Campaign</strong> from pages, files, and other content types
```
flamerohr: be careful with adding html to translation strings, it's heavily discouraged because you need to use `dangerouslySetInnerHTML` in react... (or `castStringToElement()`)
Can we use quotes instead?
I've raised this about the "Nice one!" being bold above as well, and it was agreed that we could remove the bold for that.

src/CampaignAdmin.php
```diff
@@ -155,6 +155,35 @@ public function readCampaigns()
...
+        if (class_exists('SilverStripe\CMS\Model\SiteTree')) {
+            $class = singleton('SilverStripe\CMS\Model\SiteTree');
+            $groups[] = [
+                'baseClass' => DataObject::getSchema()->baseDataClass('SilverStripe\CMS\Model\SiteTree'),
+                'singular' => $class->i18n_singular_name(),
+                'plural' => $class->i18n_plural_name(),
+                'noItemText' => 'Add from ' . $class->i18n_plural_name() . ' area',
+                'items' => []
+            ];
+        }
```
flamerohr: I'm not comfortable with the idea of cross-module references like this...
Can these be added by an extension hook, rather than hardcoded in to campaign admin?
It could even be an array of classes in a config.

Either extension hook or config will be fine, both could be a best of both worlds.

src/CampaignAdmin.php
```diff
@@ -155,6 +155,35 @@ public function readCampaigns()
...
+
+        if (class_exists('SilverStripe\AssetAdmin\Controller\AssetAdmin')) {
+            $class = singleton('SilverStripe\Assets\File');
+            $groups[] = [
+                'baseClass' => DataObject::getSchema()->baseDataClass('SilverStripe\Assets\File'),
+                'singular' => $class->i18n_singular_name(),
+                'plural' => $class->i18n_plural_name(),
+                'noItemText' => 'Add from ' . $class->i18n_plural_name() . ' area',
```
flamerohr: Oops and this could be i18n translated

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -163,29 +191,36 @@ class CampaignAdminList extends SilverStripeComponent {
...
       accordionBlocks.push(
-        <AccordionBlock key={groupid} groupid={groupid} title={title}>
-          {listGroupItems}
-        </AccordionBlock>
+        <div className={wrapperClassnames}>
+          <AccordionBlock key={groupid} groupid={groupid} title={title}>
+            {
+              listGroupItems.length > 0
+                ? listGroupItems
+                : <p className="list-group-item">{group.noItemsText}</p>
```
flamerohr: @phalkunz please note, we're following airbnb formatting standards (to an extent) so ternary conditions should break like this:
```
condition
  ? item
  : null
```
ref: https://github.com/airbnb/javascript/#comparison--nested-ternaries the better section

phalkunz: 👍

src/CampaignAdmin.php
```diff
@@ -155,6 +159,34 @@ public function readCampaigns()
         return $response;
     }
 
+    protected function getPlaceholderGroups()
```
tractorcow: PHPDoc please.

src/CampaignAdmin.php
```diff
@@ -155,6 +159,34 @@ public function readCampaigns()
...
+    {
+        $groups = [];
+
+        $classes = Config::inst()->get(self::class, 'placeholder_group_classes');
```
tractorcow: This is a duplication of ChangeSet::important_classes config.

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -315,6 +374,19 @@ class CampaignAdminList extends SilverStripeComponent {
     return groups;
   }
 
+  getPlaceholderGroups() {
```
tractorcow: I'm not sure why we need this api; doesn't groupItemsForSet() build the groups automatically?

tractorcow: ok, @phalkunz has explained why this is necessary. The only change I'd request is the switch to using the existing config for important_classes instead.

src/CampaignAdmin.php
```diff
@@ -486,7 +519,17 @@ public function getCampaignEditForm($id)
             $fields,
             FieldList::create(
                 FormAction::create('save', _t(__CLASS__.'.SAVE', 'Save'))
-                    ->setIcon('save'),
+                    ->setIcon('save')
+                    ->setSchemaState([
+                        'data' => [
+                            'pristineTitle' => _t(__CLASS__.'SAVED', 'Saved'),
+                            'pristineIcon' => 'tick',
+                            'dirtyTitle' => _t(__CLASS__.'SAVECHANGES', 'Save changes'),
```
phalkunz: @unclecheese Looks like this render the title set above (line 521) redundant.

src/CampaignAdmin.php
unclecheese: What do you mean by this? Should we be using `ChangeSet.important_classes` instead?

unclecheese: Never mind, I see your comment below.

src/CampaignAdmin.php
```diff
@@ -156,6 +158,37 @@ public function readCampaigns()
...
+     */
+    protected function getPlaceholderGroups()
+    {
+        $groups = [];
+
+        $classes = Config::inst()->get(ChangeSet::class, 'important_classes');
+        
+        foreach ($classes as $class) {
+            /** @var DataObject $item */
+            $item = Injector::inst()->get($class);
```
tractorcow: get() won't return null if the class doesn't exist, it'll throw an exception. Can you use class_exists() instead please?

PR #42, opened 2017-08-10
Upgrade bootstrap to alpha6

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-framework/issues/6380


src/CampaignAdmin.php
```diff
@@ -527,7 +527,7 @@ public function getCampaignEditForm($id)
                             'pristineIcon' => 'tick',
                             'dirtyTitle' => _t(__CLASS__.'SAVE', 'Save'),
                             'dirtyIcon' => 'save',
-                            'pristineClass' => 'btn-primary-outline',
+                            'pristineClass' => 'btn-outline-outline',
```
flamerohr: this should be `btn-outline-primary` I think

tractorcow:
Fixed.

Issue #39, opened 2017-07-24
Remove item from campaign within campaigns UI (incl. dependencies)


clarkepaul:
## Problem
There is no way to remove pages/files from a campaign as yet, this is a pretty fundamental part of campaigns especially if you want to publish the campaign but don't want to publish a single item.

## AC's
* As a user, I am presented with an option to remove an item from a campaign
* If the item I want to remove doesn't have any dependencies then the item can be removed
* If the item has dependencies then I am informed that to remove this item I must also remove all dependencies
* If a selected item is removed (user should only be able to remove selected items as part of this issue) the item and preview disappear and the user is taken to the next item in the list if available
* Files that can be removed are removed instantly (there is no "Undo" functionality, nor any confirmation message)

## Excluded
* As a user I can opt to protect/safeguard an item from being published as part the campaign regardless of whether an item is explicitly added or implicitly added (see ...)
* As a user, I have the option to remove a parent without removing its children

## Designs
Aspiration of design direction (we will only build a sub-set of this functionality as part of this issue)
https://invis.io/DBCXO172C#/247538915_Campaigns_Camapgin_Detail_Preview_Item

Differences:
* Files that can be removed are removed instantly (there is no "Undo" functionality, nor any confirmation message)
* We are only implementing the "Remove from campaign" action within the more options drop-down on selected item preview panel
<img width="168" alt="pasted_image_23_08_17__1_58_pm" src="https://user-images.githubusercontent.com/555033/29595383-39ee13fc-880b-11e7-8ec6-a56e0b811f3a.png">
* If a selected item is removed (user should only be able to remove selected items as part of this issue) the item and preview disappear and the user is taken to the next item in the list if available. 

## Tasks

 - [x] Add remove button if item can be removed
 - [x] Add message if implicit item cannot be removed
 - [x] Confirmation before removing
 - [x] Confirm UX with @clarkepaul 
 
## Pull requests

- [x] https://github.com/silverstripe/silverstripe-admin/pull/207
- [ ] https://github.com/silverstripe/silverstripe-campaign-admin/pull/48

clarkepaul:
I'm working on the UI design for this

clarkepaul:
**Thinking out aloud here for notes**
* if a campaign contains a page (P1) and that page contains two images (F1 and F2)
* if a user chooses to remove P1 offer them two options:
   * remove all three items will be removed P1, F1, F2 (if not used elsewhere)
[...]

PR #48, opened 2017-08-28
Add remove action to campaign item

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-campaign-admin/issues/39


dhensby:
needs rebasing/building

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -206,13 +218,49 @@ class CampaignAdmin extends SilverStripeComponent {
...
+By removing this item all linked items will be removed unless used elsewhere.`;
+    const msg = i18n._t('CampaignAdmin.REMOVE_ITEM_MESSAGE', fallbackMsg);
+    // eslint-disable-next-line no-alert
+    const confirmed = window.confirm(msg);
+
+    if (!confirmed) {
+      return false;
+    }
+
+    this.props.campaignActions.removeCampaignItem(
```
flamerohr: you should return this promise, rather than an empty `true`

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -206,13 +218,49 @@ class CampaignAdmin extends SilverStripeComponent {
...
+      this.removeCampaignItemApi,
+      campaignId,
+      itemId
+    )
+      .then(() => {
+        const endpoint = this.props.sectionConfig.readCampaignsEndpoint;
+        const fetchURL = endpoint.url;
+        this.props.recordActions.fetchRecords(
```
flamerohr: you can return this inside the `then`

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -206,13 +218,49 @@ class CampaignAdmin extends SilverStripeComponent {
...
+          this.props.sectionConfig.treeClass,
+          endpoint.method,
+          fetchURL
+        )
+          .then(() => {
```
flamerohr: after returning `this.props.recordActions.fetchRecords(` you can move this to outside the parent `then`

So you'll have:
```
return this.props.campaignActions.removeCampaignItem()
.then(return fetchRecords)
.then(selectChangeSetItem());
```

Welcome to promise chains!

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -206,13 +218,49 @@ class CampaignAdmin extends SilverStripeComponent {
...
+            this.props.campaignActions.selectChangeSetItem(null);
+            // Workaround to hide more actions popover
+            window.document.body.click();
+          });
+      });
+
+    return true;
```
flamerohr: So the reason to return the promise, rather than this empty `true` is for unit testing...

I would like to see a unit test covering the functions that need to be called in this chain.

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -99,8 +100,12 @@ class CampaignAdminList extends SilverStripeComponent {
       selected = items.find(item => itemId === item.ID);
     }
 
-    if (!selected && items.length > 0) {
-      selected = items[0];
+    if (!selected) {
+      // Select the first item in the first display group
+      const groupItems = this.groupItemsForSet();
+      const groupItemsKeys = Object.keys(groupItems);
+      const group = groupItemsKeys.length > 0 ? groupItems[groupItemsKeys[0]] : null;
```
flamerohr: move `groupItemsKeys[0]` to a variable please

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -274,13 +279,69 @@ class CampaignAdminList extends SilverStripeComponent {
...
+      selectedItem._links.referenced_by.length || 0;
+    const unremoveableInfoText = requiredByNum < 2 ?
+      i18n._t(
+        'CampaignAdmin.UNREMOVEABLE_INFO_SINGULAR',
+        'Required by %s item, and cannot be removed directly.'
+      ) :
+      i18n._t(
+        'CampaignAdmin.UNREMOVEABLE_INFO_PLURAL',
+        'Required by %s items, and cannot be removed directly.'
+      );
```
flamerohr: I strongly don't agree with having plurals hardcoded like this... we've talked about this before as well. This doesn't translate well for languages with different pluralisation rules to english (like Russian) and duplication work for no pluralisation rules (like Chinese).

Please change this to `'Required by {number} item(s), and cannot be removed directly.'`

If you'd like we can work on (or raise a story for) a pluralisation API for the javascript side, which can be similar to what was implemented on the PHP side.

Missing entry for new translation strings in `en.js` and `en.json`

src/CampaignAdmin.php
```diff
@@ -396,6 +406,42 @@ public function readCampaign(HTTPRequest $request)
...
+            return new HTTPResponse(null, 400);
+        }
+
+        $campaignID = $request->param('CampaignID');
+        $itemID = $request->param('ItemID');
+
+        if (!$campaignID ||
+            !is_numeric($campaignID) ||
+            !$itemID ||
+            !is_numeric($itemID)) {
```
flamerohr: having `!$campaignID` and `!$itemID` seems redundant, if an invalid numeric ID is provided (`0`) it should return a `404` error

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -274,13 +279,69 @@ class CampaignAdminList extends SilverStripeComponent {
...
+          className="btn btn-secondary action"
+          onClick={this.handleRemoveItem}
+        >
+          Remove
+        </button>
+      ) :
+      (
+        <p key="unremoveable_info" className="alert alert-info campaign-admin__unremoveable-item">
+          <span className="font-icon-link"></span>
+          {i18n.sprintf(unremoveableInfoText, requiredByNum)}
```
flamerohr: Please use i18n.inject() with named attributes instead of i18n.sprintf(), so we don't have positional issues later on :)

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -206,13 +218,49 @@ class CampaignAdmin extends SilverStripeComponent {
...
     return (
       <CampaignAdminList {...props} />
     );
   }
 
+  handleRemoveCampaignItem(campaignId, itemId) {
+    const fallbackMsg = `Are you sure you want to remove this item?
+
+By removing this item all linked items will be removed unless used elsewhere.`;
+    const msg = i18n._t('CampaignAdmin.REMOVE_ITEM_MESSAGE', fallbackMsg);
```
flamerohr: Missing `en.js` and `en.json` entry for translation strings

client/src/containers/CampaignAdmin/CampaignAdminList.js
phalkunz: Sorry, I totally forgot about this. I've created an issue for the API (https://github.com/silverstripe/silverstripe-admin/issues/218).

client/src/state/campaign/tests/CampaignActions-test.js
```diff
@@ -0,0 +1,62 @@
...
+describe('CampaignActions', () => {
+  const middlewares = [thunk];
+  const mockStore = configureMockStore(middlewares);
+
+  describe('removeCampaignItems()', () => {
+    it('should create 2 actions, request and success, when resolved', () => {
+      const store = mockStore({});
+      const removeItemApi = () => new Promise((resolve) => {
+        resolve();
+      });
```
flamerohr: you can shorten this to `const removeItemApi = () => Promise.resolve();` :)

flamerohr:
Looks good, can be merged on green

PR #52, opened 2017-08-31
FIX crop campaign item header text

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/5459


unclecheese:
Merge on green.

PR #58, opened 2017-09-06
ENHANCEMENT Don’t automatically show inferred campaigns

tractorcow:
API Allow campaign visibility to be configured by type via config
ENHANCEMENT Minimise excessive sync behaviour

Requires https://github.com/silverstripe/silverstripe-versioned/pull/44


tractorcow:
Before:

image

After:

image

PR #64, opened 2017-09-18
Fix add null check for lastSynced

flamerohr:


chillu:
Damian peer pressured me into merging this

PR #71, opened 2017-10-19
Add behat tests

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/6265


behat.yml
```diff
@@ -0,0 +1,34 @@
+# Run asset-admin behat tests with this command (installed with silverstripe/installer)
+# Note that asset-admin behat tests require CMS module
+# ========================================================================= #
+# vendor/bin/selenium-server-standalone -Dwebdriver.firefox.bin="/Applications/Firefox31.app/Contents/MacOS/firefox-bin"
+# vendor/bin/serve --bootstrap-file vendor/silverstripe/cms/tests/behat/serve-bootstrap.php
+# vendor/bin/behat @asset-admin
```
tractorcow: Forgot to change asset-admin to campaign-admin :)

behat.yml
```diff
@@ -0,0 +1,34 @@
...
+        - SilverStripe\BehatExtension\Context\LoginContext
+        -
+          SilverStripe\CampaignAdmin\Tests\Behat\Context\FixtureContext:
+            - %paths.modules.campaign-admin%/tests/behat/files/
+        -
+          SilverStripe\Framework\Tests\Behaviour\ConfigContext:
+            - %paths.modules.campaign-admin%/tests/behat/files/
+  extensions:
+    SilverStripe\BehatExtension\Extension:
+      bootstrap_file: vendor/silverstripe/cms/tests/behat/serve-bootstrap.php
```
tractorcow: Since we're using cms here let's add`cms` to the require-dev (same as asset-admin does). This is sufficient for dependencies only required during testing.

tests/behat/features/populate-campaigns.feature
```diff
@@ -0,0 +1,88 @@
...
+    Then I should see "Successfully added" in the ".modal-dialog" region
+    When I close the modal
+    And I go to "/admin/campaigns"
+    And I wait until I see the ".grid-field__table" element
+    Then I should see "2 total (2 changes)" in column 2 of the "Empty Campaign" campaign
+    When I view the campaign "Empty Campaign"
+    Then I should see the "About us changedbout Us" campaign item
+    And I should see the "renamedfile" campaign item
+    And the "Publish campaign" button should not be disabled
+Scenario: I can manage campaign items
```
tractorcow: Indentation is off here.

tests/behat/features/populate-campaigns.feature
```diff
@@ -0,0 +1,88 @@
...
+    When I click the ".notice-item-close" element
+    And I click "More options" in the "#ActionMenus" element
+    And I press the "Add to Campaign" button
+    Then I should see a modal titled "Add to campaign"
+    When I select "Empty Campaign" from "Campaign"
+    And I press the "Add" button
+    And I wait for 1 seconds until I see the ".modal__response--good" element
+    Then I should see "Successfully added" in the ".modal-dialog" region
+    When I close the modal
+    And I go to "/admin/assets"
```
tractorcow: Also note to add asset-admin to require-dev

tests/behat/src/FixtureContext.php
```diff
@@ -0,0 +1,281 @@
...
+        assertNotNull($button, 'Close button not found');
+
+        $button->click();
+    }
+
+    /**
+     * @Given /^I click the "([^"]+)" gallery item$/
+     * @param $name
+     */
+    public function iClickTheGalleryItem($name)
```
tractorcow: Add the AssetAdmin version of FixtureContext to your behat.yml as well, and use stepISelectGalleryItem() instead of duplicating.

tests/behat/src/FixtureContext.php
```diff
@@ -0,0 +1,281 @@
...
+            }
+            $campaign->addObject($record);
+        }
+    }
+
+    /**
+     * @When /^I click the "([^"]+)" element$/
+     * @param $selector
+     */
+    public function iClickTheElement($selector)
```
tractorcow: Duplicate of `SilverStripe\BehatExtension\Context\BasicContext::iClickInTheElement()`

tests/behat/src/FixtureContext.php
```diff
@@ -0,0 +1,281 @@
...
+        /*@ var Behat\Mink\Driver\Selenium2Driver $driver */
+        $driver = $this->getMainContext()->getSession()->getDriver();
+        $driver->getWebDriverSession()->accept_alert();
+    }
+
+    /**
+     * @Then /^I should see the "([^"]*)" form$/
+     * @param string $id HTML ID of form
+     */
+    public function iShouldSeeTheForm($id)
```
tractorcow: Can use `SilverStripe\AssetAdmin\Tests\Behat\Context\FixtureContext::iShouldSeeTheForm` instead of duplicating, after adding asset admin to require-dev.

tests/behat/src/FixtureContext.php
unclecheese: There was nothing I could find for clicking a specific element by selector. `iClickInTheElement()` looks for text in an element, which doesn't always work. For instance, a `<button />` that's just an empty tag with an icon in a `:before` pseudo class.

tests/behat/src/FixtureContext.php
unclecheese: I tried this, but it actually doesn't work, because they both inherit from `SilverStripe\BehatExtension\Context\FixtureContext`, so there are all kinds of collisions on the inherited methods, e.g:

```
      SilverStripe\CampaignAdmin\Tests\Behat\Context\FixtureContext::assertMessageBoxContainsText()
      SilverStripe\AssetAdmin\Tests\Behat\Context\FixtureContext::assertMessageBoxContainsText()
    And I click "More options" in the "#ActionMenus" element
      Step "/^I should see "(?P<text>(?:[^"]|\\")*)" in the message box$/" is already defined in SilverStripe\CampaignAdmin\Tests\Behat\Context\FixtureContext::assertMessageBoxContainsText()
```

tests/behat/src/FixtureContext.php
unclecheese: Same as above. Can't use two descendants of the base `FixtureContext`

tractorcow: Make your new one extend the prior instead?

tests/behat/src/FixtureContext.php
tractorcow: If that's the case, I suggest to add it to that context. This is an extremely generic action, but on a very specific use case context.

tests/behat/src/FixtureContext.php
tractorcow: We decided to go with the inline copy for now; It's not that big a cost to have a minor overlap.  :)

tractorcow:
Can be merged on green

PR #72, opened 2017-10-31
Feature a help panel to describe what campaigns are for

flamerohr:
Also, it is extensible for later to add links to userhelp docs

screen shot 2017-10-31 at 3 29 24 pm
screen shot 2017-10-31 at 3 29 34 pm

*NOTE*: the links and buttons do not show in the pull request, they're there for preview purposes

Requires https://github.com/silverstripe/silverstripe-admin/pull/333 for new icons added


tractorcow:
Looks good, but please see my comment on the admin pr.

client/src/state/campaign/CampaignReducer.js
```diff
@@ -1,12 +1,16 @@
...
 
+// retrieve showMessage state from user's localStorage
+const showMessage = window.localStorage.getItem('campaign.showMessage');
 const initialState = deepFreeze({
   campaignId: null,
   changeSetItemId: null,
   isPublishing: false,
   view: null,
   newItem: null,
+  showMessage: showMessage === null,
```
tractorcow: We should only show message if showMessage == null && there are no campaigns; I have campaigns loaded, and my initial load displayed the help message when I didn't expect it.

tractorcow:
On second thoughts, I don't think it's that important. Let's just get this in as is (once we fix the icons).

client/src/state/campaign/CampaignReducer.js
flamerohr: To elaborate:
The trouble with adding the logic to check for campaigns - the `showMessage` state becomes ternary and ambiguous until the list of Campaign records has been received, meaning a few extra layers for something that is stored in `localStorage` after the first time it's closed anyway.

P.S. Great that we agree that it's not that important.

newleeland:
Could we add more padding? (padding: 32px 30px 30px 30px;)
screen shot 2017-11-01 at 2 19 36 pm


Could we remove the right margin on mobile view?
screen shot 2017-11-01 at 1 43 53 pm

flamerohr:
Padding and margins amended

Issue #50, opened 2017-08-30
Add a confirm action for publishing a campaign as a precaution


clarkepaul:
Unintentionally publishing a campaign could be a big mistake if the publish action was accidentally clicked on, we should probably have a confirmation action/dialog. Normally I would prefer an "Undo" action instead of a confirm but that is in the "too hard" basket, based on conversations with the SS OS team.

## Pull request

- [x] https://github.com/silverstripe/silverstripe-campaign-admin/pull/80

PR #80, opened 2017-11-22
Display confirmation before publishing a compaign

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-campaign-admin/issues/50


tractorcow:
Looks good, but need to get CI passing. I've added a tweak that may fix it.

PR #82, opened 2017-12-10
Update variable name to new name

flamerohr:


flamerohr:
All tests passing except for behat which is broken due to FirefoxDrivers (not due to the pull request)

PR #85, opened 2018-01-19
Remove references to old Selenium2Driver

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-behat-extension/issues/137


tractorcow:
Note, don't merge this without us agreeing to shift entirely away from selenium. It's a bit of a large job to migrate everything.

PR #90, opened 2018-02-26
Enhancement removed max width for content in intro screen

flamerohr:
I don't see why it's needed yet :) (and I'm the original author of it)

Re-discovered it as part of https://github.com/silverstripe/silverstripe-admin/issues/417


unclecheese:
Any idea why it hosed the Behat test? Doesn't seem to be related.

PR #95, opened 2018-04-17
FIX Re-implement flexbox classes for preview window

robbieaverill:
This is a regression from https://github.com/silverstripe/silverstripe-admin/pull/478

Re-adding the required flexbox classes 

Before:
image

After:
image


flamerohr:
behat broken, seems related for some reason...

robbieaverill:
I think it's coincidence =D the latest 1 branch build has the same failure: https://travis-ci.org/silverstripe/silverstripe-campaign-admin/jobs/365852378

flamerohr:
Merging on the understanding that the break is pre-existing ;)

PR #96, opened 2018-04-17
MINOR Use more object destructuring and spread syntax

robbieaverill:
Uses destructuring more in props, and replaces Object.assign with spread syntax

Also fixes #96


robbieaverill:
Ping me when reviewing and I'll rebase

robbieaverill:
@maxime-rainville updated, I've also added the missing translation defaults to en.json, added one which was not translated yet and switched the two sprintf style translations which weren't in the source dictionary yet to use inject instead while I was there

PR #108, opened 2018-07-01
FIX Button outline secondary class is now correct in disabled "Publish campaign" button

robbieaverill:
Before:
image

Should be/is now:
image


robbieaverill:
I've updated the pattern library and docs for FormAction at https://github.com/silverstripe/silverstripe-admin/pull/557

robbieaverill:
Rebased, cc @clarkepaul @ScopeyNZ could you please review?

ScopeyNZ:
LGTM. I'll merge and we can follow up if @clarkepaul thinks the old way was better 🤢

PR #109, opened 2018-07-11
NEW Make use of ViewModeToggle component.

raissanorth:
Allows users to toggle between edit, preview, and split mode in the CMS. 

Introduces dependencies on `react-resize-aware` library and `Enzyme` testing utility. The former is used to retrieve the screen width and determine if split mode should be enabled based on it. The latter is used to shallow render `CampaignAdminList` in the unit tests, as we don't want to test the child components `ViewModeToggle` and `Preview`. These components are now provided via Injector.

Removed media queries in favour of rendering preview and/or edit based on the state stored in `ViewModeToggle`'s store.

Refactored `CampaignAdminList`'s render method, to make it more readable. Moved the logic to render the detail/edit view into its own method - `renderCampaignAdminListDetail`


client/src/containers/CampaignAdmin/CampaignAdmin.scss
```diff
@@ -3,33 +3,27 @@
...
+    //  @include media-breakpoint-down(lg) {
+    //    border: 0;
+    //  }
+    //
+    //  &:hover {
+    //    background-color: $list-group-hover-bg;
+    //    color: $list-group-link-color;
+    //    border-color: $list-group-border-color;
+    //  }
+    //}
```
robbieaverill: We can remove this if it's not required any more

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -536,6 +545,10 @@ function mapDispatchToProps(dispatch) {
     breadcrumbsActions: bindActionCreators(breadcrumbsActions, dispatch),
     campaignActions: bindActionCreators(campaignActions, dispatch),
     recordActions: bindActionCreators(recordActions, dispatch),
+
+    onResize(panelWidth) {
```
robbieaverill: Could you add a proptype for this please?

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -279,53 +293,114 @@ class CampaignAdminList extends Component {
...
     );
   }
 
   renderPreview(itemLinks, itemId) {
-    let preview = null;
     const previewClasses = classnames([
       'flexbox-area-grow',
       'fill-height',
       'preview',
-      'campaign-admin__campaign-preview',
```
robbieaverill: Any reason this was removed?

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -279,53 +293,114 @@ class CampaignAdminList extends Component {
...
+      />
+    );
+  }
+
+  /**
+   * Renders the details section of the campaign list.
+   *
+   * @param body
+   *
+   * @return object
```
robbieaverill: I think these JS doc annotations are a little out

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -279,53 +293,114 @@ class CampaignAdminList extends Component {
...
+    // Hide when the preview mode is explicitly enabled
+    if (previewState === 'preview') {
+      return null;
+    }
+
+    let classNames = 'fill-height campaign-admin__campaign-items';
+    // Modify width when the edit mode is enabled
+    if (previewState === 'edit') {
+      classNames = 'fill-height campaign-admin__campaign-items-edit';
+    }
```
robbieaverill: Using the classnames lib here might be a little tidier

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -279,53 +293,114 @@ class CampaignAdminList extends Component {
...
+
+    return (
+      <div className={classNames} aria-expanded="true">
+        <Toolbar showBackButton onBackButtonClick={onBackButtonClick}>
+          <Breadcrumb multiline />
+        </Toolbar>
+        {newItemInfo}
+        <div className={bodyClass.join(' ')}>
```
robbieaverill: Yeah again use classnames

raissanorth:
Note that this is dependant on https://github.com/silverstripe/silverstripe-admin/pull/560

client/src/containers/CampaignAdmin/CampaignAdminList.js
raissanorth: This class was used to hide the preview on smaller screens. Media queries are no longer used for this behaviour, but the ViewModeToggle's state. Therefore the class is redundant.

robbieaverill: Sure, but removing it is kind of a breaking change - you've already removed the styles for it, but someone else might be relying on it too

raissanorth: That makes sense, and is something I haven't had considered. I will revert my change.

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -252,7 +253,20 @@ class CampaignAdminList extends Component {
...
+  renderViewModeToggle() {
+    const { ViewModeComponent } = this.props;
+
+    return (
+      <ViewModeComponent
+        id="view-mode-toggle-in-edit-nb"
+        area={'edit'}
+      />
+    );
+  }
```
robbieaverill: It looks like this method isn't used at all - perhaps you could remove it. The component is rendered in the `render()` method directly and doesn't need to be found to the instance in construction I don't think

flamerohr:
@robbieaverill make sure to bump the version dependency in the package.json as well

also, I'm not sure why tests are failing for `react-apollo`

robbieaverill:
>also, I'm not sure why tests are failing for react-apollo

@flamerohr FYI this was because the admin module npm dependencies weren't installed. I reproduced this on the latest 1 branch for campaign-admin in Travis, we've pushed a fix for it which is stolen from asset-admin

robbieaverill:
Merge on green

flamerohr:
oh I see *cringe* lol

PR #110, opened 2018-07-15
MINOR Fix close and help button appearance

### LINKED ISSUE #86: Help icon is not centered horizontally
Add the class .btn--no-text to the button so that it removes the space which typically sits between the icon and text.

image

lukereative:
Fixes #86


client/src/components/IntroScreen/IntroScreen.js
```diff
@@ -50,7 +50,7 @@ class IntroScreen extends Component {
     return (
       <div className="fill-width campaign-info">
         <div className="campaign-info__icon">
-          <span className="font-icon-white-question icon btn--icon-xl" />
+          <span className="font-icon-white-question icon btn--icon-xl btn--no-text" />
```
maxime-rainville: We should have an aria label on this.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -416,7 +416,7 @@ By removing this item all linked items will be removed unless used elsewhere.`;
                   role="button"
                   tabIndex={0}
                   onClick={this.handleToggleMessage}
-                  className="btn btn-secondary font-icon-white-question btn--icon-xl"
+                  className="btn btn-secondary font-icon-white-question btn--icon-xl btn--no-text"
```
maxime-rainville: We should have an aria label on this.

PR #112, opened 2018-08-01
FIX Toolbar button margins are constrained to campaign previews, and update ViewModeActions name

robbieaverill:
This CSS style was too broad, so updated to be specific to campaign previews

Also ViewModeActions name changed recently, so updating that too.

Before:
image
image

After:
image

image


maxime-rainville:
I presume it's not normal for me to loose all campaign actions when I switch to the Edit mode?

https://youtu.be/cAOkDXFGAgE

robbieaverill:
You're right, but it sounds like a pre-existing bug to me. I'll look at it separate from this PR if that's OK

robbieaverill:
Hmm @maxime-rainville, you don't have any actions to start with so you aren't really losing any... I can't reproduce your issue.

I've pushed another fix to this branch and have logged https://github.com/silverstripe/silverstripe-campaign-admin/issues/113 for further follow up bugs related to this work

PR #114, opened 2018-08-15
FIX text/json is not a valid mimetype

dhensby:
Partial fix of https://github.com/silverstripe/silverstripe-framework/issues/5594


kinglozzer:
Ah 💩, I merged this before spotting @robbieaverill’s [comment](https://github.com/silverstripe/silverstripe-framework/pull/8327#issuecomment-413307707) that this probably belongs in a patch release... should we cherry pick this commit down to the minor release branches?

robbieaverill:
Up to you guys, I don’t mind :-)

PR #116, opened 2018-08-31
BUG: Do not render campaign toolbar if there are no items in it yet

bergice:


client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -386,6 +386,9 @@ class CampaignAdminList extends Component {
       }
     );
 
+    let items = this.getItems();
```
ScopeyNZ: In this case `items` can be a `const` too 🙂

dhensby:
Updated, but I think that it's likely the behat test will fail again - I think something isn't quite working right here?

PR #136, opened 2019-02-10
FIX Implicit changeset items should be able to opt out of UI display in campaign admin

robbieaverill:
Implicitly added changes (via an ORM relationship or similar) are automatically shown in campaign views. There should have been a way for individual models to opt out of this if for example they are link models or something else that should be invisible to CMS users.

This PR adds configuration checks for opt out in order for this to be possible.

Issue: https://github.com/dnadesign/silverstripe-elemental/issues/593

### DIFF
--- a/src/CampaignAdmin.php
+++ b/src/CampaignAdmin.php
@@ -308,7 +308,9 @@ class CampaignAdmin extends LeftAndMain implements PermissionProvider
 
                 /** @var ChangesetItem $changeSetItem */
                 $resource = $this->getChangeSetItemResource($changeSetItem);
-                $hal['_embedded']['items'][] = $resource;
+                if ($resource) {
+                    $hal['_embedded']['items'][] = $resource;
+                }
             }
 
         // An unexpected data exception means that the database is corrupt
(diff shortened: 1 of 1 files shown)


src/CampaignAdmin.php
```diff
@@ -324,12 +326,18 @@ protected function getChangeSetResource(ChangeSet $changeSet, $sync = false)
...
+     * @return array|false
      */
     protected function getChangeSetItemResource(ChangeSetItem $changeSetItem)
     {
         $baseClass = DataObject::getSchema()->baseDataClass($changeSetItem->ObjectClass);
         $baseSingleton = DataObject::singleton($baseClass);
+
+        // Allow items to opt out of being displayed in changesets
+        if ($baseSingleton->config()->get('hide_in_campaigns')) {
+            return false;
```
ScopeyNZ: Is this safe? Would `return []` have the same affect without changing the return type?

robbieaverill: We could do either - I went for false because it's more descriptive, and this method isn't part of our public API so changing its return type doesn't matter as much, but you're probably right that returning an empty array is safer. Will update.

PR #156, opened 2020-02-10
FIX Update core requirements to 4.5 series

Cheddam:
See [silverstripe-cms#2521](https://github.com/silverstripe/silverstripe-cms/issues/2521) for context.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,15 +21,15 @@
         }
     ],
     "require": {
-        "silverstripe/admin": "^1@dev",
-        "silverstripe/framework": "^4@dev",
-        "silverstripe/versioned": "^1@dev",
+        "silverstripe/admin": "^1.5",
+        "silverstripe/framework": "^4.5",
+        "silverstripe/versioned": "^1.5",
         "silverstripe/vendor-plugin": "^1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",
-        "silverstripe/asset-admin": "^1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -21,15 +21,15 @@
         }
     ],
     "require": {
-        "silverstripe/admin": "^1@dev",
-        "silverstripe/framework": "^4@dev",
-        "silverstripe/versioned": "^1@dev",
+        "silverstripe/admin": "^1.5",
```
emteknetnz:
```suggestion
        "php": ">=7.1",
        "silverstripe/admin": "^1.5",
```

robbieaverill: I don't agree with introducing this - we already list PHP requirements in framework, that's a single place where we need to maintain it at the moment

emteknetnz: Unfortunately that approach doesn't always work, it is what led to the original creation of this issue https://github.com/silverstripe/silverstripe-cms/issues/2521

What happened is that a travis build failed because there was a module (I forget which one) that requires silverstripe/cms but not silverstripe/framework.  That particular travis test was running php 5.6 and failed because it included a version of silverstripe/cms that required php 7.1

robbieaverill: If you're adding the PHP constraint then you don't need to bump the admin constraint

emteknetnz: Apologies, I may have misinterpreted the solution.  My suggestion for adding the PHP constraint is not required as the lock-stepping takes care of the original issue.

PR #171, opened 2020-08-14
FIX Show Untitled when blank title

emteknetnz:
Fix for https://github.com/dnadesign/silverstripe-elemental/issues/574

Will show "Untitled" for any item that has a blank title, not just content blocks

Before
image

After
image


ScopeyNZ:
There's a bit of discussion in the related issue. I personally feel like we should be consistent with elemental here, so the title should be "Untitled Content block" or whatever it is in elemental. Not quite sure how we can put this into code though.

I like the idea of elemental providing a `BlockTitle` component or something, but I'm not sure how that would tie into campaign admin...

emteknetnz:
Agree it would be better, though it'll also massively increase the effort required.  

There's already context within campaign admin that the thing that is "Untitled" is a "Block"

I'd have a strong preference to keep this PR as is because it still makes things significantly better than they are now.

PR #177, opened 2020-09-16
BUG Display a notification after publishing a campaign

### LINKED ISSUE #131: User should receive a toast confirmation when publishing a campaign
When a campaign is published there is no feedback for the user to confirm that their publishing action was successful.

A toast message like 'Published "[Campaign name]" successfully.' should be used

### ACs
- Finalise existing PR
- Test and confirm functionality with PUX

## PR
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/177

maxime-rainville:
Use the new react toast notification to tell the user when their campaign has been published.

# Parent issue
* resolves #131 

# Depends on
* https://github.com/silverstripe/silverstripe-admin/pull/1068
* https://github.com/silverstripe/silverstripe-framework/pull/9770


maxime-rainville:
Might need to wrtie some unit test and behat test + bump dependencies.

client/src/state/campaign/CampaignActions.js
```diff
@@ -64,12 +66,22 @@ export function publishCampaign(publishApi, recordType, campaignId) {
           type: RECORD_ACTION_TYPES.FETCH_RECORD_SUCCESS,
           payload: { recordType, data },
         });
+        const message = i18n._t(
+          'CampaignAdmin.PUBLISH_SUCCESS',
+          '"%s" campaign has been published.'
```
sachajudd, line 71: We need to ensure we are using the same patterns in the CMS, can we check what pages do? As this is not the same wording that was recommended by Jared.

client/lang/src/en.json
```diff
@@ -8,5 +8,7 @@
     "CampaignAdmin.NO_CHANGES": "No changes",
     "CampaignAdmin.PUBLISHCAMPAIGN": "Publish campaign",
     "CampaignAdmin.REMOVED": "Removed",
-    "CampaignAdmin.REVERTCAMPAIGN": "Revert"
-}
\ No newline at end of file
+    "CampaignAdmin.REVERTCAMPAIGN": "Revert",
+    "CampaignAdmin.PUBLISH_FAIL": "Campaign could not be published.",
+    "CampaignAdmin.PUBLISH_SUCCESS": "\"%s\" campaign has been published."
```
bergice, line 13:
```suggestion
    "CampaignAdmin.PUBLISH_SUCCESS": "Published \"%s\" successfully."
```

client/src/state/campaign/CampaignActions.js
bergice, line 71:
```suggestion
          'Published "%s" successfully.'
```

bergice (changes requested):

bergice:
Seems to work, just need to update the text and fix travis.

maxime-rainville:
Updated the messages and added some behat test. The behat test won't be green until the matching framework PR is merged.

client/src/state/campaign/CampaignActions.js
maxime-rainville: Translation is now `Published \"%s\" successfully.` which is the same message used when publishing a page.

client/lang/src/en.json
```diff
@@ -8,5 +8,7 @@
     "CampaignAdmin.NO_CHANGES": "No changes",
     "CampaignAdmin.PUBLISHCAMPAIGN": "Publish campaign",
     "CampaignAdmin.REMOVED": "Removed",
-    "CampaignAdmin.REVERTCAMPAIGN": "Revert"
-}
\ No newline at end of file
+    "CampaignAdmin.REVERTCAMPAIGN": "Revert",
+    "CampaignAdmin.PUBLISH_FAIL": "Campaign could not be published.",
```
sachajudd, line 12: @maxime-rainville What does a page do when it fails publishing?

client/lang/src/en.json
```diff
@@ -8,5 +8,7 @@
     "CampaignAdmin.NO_CHANGES": "No changes",
     "CampaignAdmin.PUBLISHCAMPAIGN": "Publish campaign",
     "CampaignAdmin.REMOVED": "Removed",
-    "CampaignAdmin.REVERTCAMPAIGN": "Revert"
-}
\ No newline at end of file
+    "CampaignAdmin.REVERTCAMPAIGN": "Revert",
```
sachajudd, line 11: Something like: Reverted \"%s\" could give more context as to what is being reverted?

maxime-rainville: As far as I know, that feature hasn't been implemented.

client/lang/src/en.json
maxime-rainville: The expectation here is that the back end will give us a proper error message and we will display that.

This string is only here in case we don't get an error message back and we don't know why publication failed.

sachajudd (approved):
Approved from UX perspective.

emteknetnz (changes requested):
Code looks fine and tests fine

It might be worth creating another issue for making a toast notification for when you inline CREATE a campaign 

image

You've got a behat timeout on travis that'll need to be fixed

maxime-rainville:
Just had a glance to see if we could sneak in a "successfully added" toast notification is this PR. Sadly, it looks non-trivial. 

The add to campaign logic is actually in admin for some weird reason. https://github.com/silverstripe/silverstripe-admin/blob/1/client/src/legacy/AddToCampaignForm.js

maxime-rainville:
Added a follow up issue for "the add to campaign" modal https://github.com/silverstripe/silverstripe-campaign-admin/issues/190

PR #178, opened 2020-10-02
FIX Address regression caused by reactstrap upgrade

emteknetnz:
Part of https://github.com/silverstripe/silverstripe-admin/issues/1126


dnsl48 (approved):
looks good

PR #233, opened 2022-05-12
ENH Added cascade_add_to_campaign property

tiller1010:
### Summary
Added `$cascade_addtocampaign` property. This explicitly adds related objects to a ChangeSet when "Add to campaign" is clicked. This allows us to bulk publish child objects without defining them in `$owns` (sometimes we want to publish data on the parent object, without publishing all child objects too).

```
class Category extends DataObject {

  private static $db = [
    "Description" => "HTMLText", // We may want to change and publish just this Content, without effecting Products
  ];

  private static $many_many = [
    'Products' => Product::class,
  ];
  
  // Allows us to bulk publish Products in a Category
  private static $cascade_addtocampaign = [
    "Products",
  ];

}
```

### Testing
- [ ] define a `$cascade_addtocampaign` property on a parent DataObject like the setup above
- [ ] add a parent object with children to a campaign
- [ ] confirm the relationships are added to the same campaign

### Notes
- We require this functionality for a site running on php 7.3, so this is set to merge into the 1.9 branch.

### DIFF
--- a/src/AddToCampaignHandler.php
+++ b/src/AddToCampaignHandler.php
@@ -317,6 +317,13 @@ class AddToCampaignHandler
 
         $changeSet->addObject($object);
 
+        $childObjects = $object->findRelatedObjects('cascade_add_to_campaign');
+        if ($childObjects) {
+            foreach ($childObjects as $childObject) {
+                $changeSet->addObject($childObject);
+            }
+        }
+
         $request = $this->controller->getRequest();
         $message = _t(
             __CLASS__ . '.Success',


src/AddToCampaignHandler.php
```diff
@@ -317,6 +317,13 @@ public function addToCampaign($object, $data)
 
         $changeSet->addObject($object);
 
+        $childObjects = $object->findRelatedObjects('cascade_addtocampaign', true);
```
GuySartorelli, line 320:
```suggestion
        $childObjects = $object->findRelatedObjects('cascade_add_to_campaign');
```
Make the name of the config more consisten with other config variables.
And we don't need to pass true as the second argument, it's true by default.

GuySartorelli (changes requested):
Please remove the extension - we don't need to set an empty array since `findRelatedObjects` will do that for us if it's falsy.

GuySartorelli:
Thanks for making those changes.
Did you rebase when you retargetted the PR? (Sorry, I should have included that in my comment) - the travis tests seem to be using an old configuration which is causing them to fail, instead of picking up the correct configuration from the `1` branch. If you didn't rebase, can you please do that?

If you could also squash your commits down to being a single commit that would be helpful, though I can squash on merge if you have issues doing that (I know squashing can get tricky sometimes).

tiller1010:
@GuySartorelli Let me know how this looks!

GuySartorelli (approved):

GuySartorelli:
Looks good, thanks for the contribution.

PR #244, opened 2022-10-11
API Strongly-type action method signatures

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10466

### DIFF
--- a/src/AddToCampaignHandler.php
+++ b/src/AddToCampaignHandler.php
@@ -285,10 +285,9 @@ class AddToCampaignHandler
      *
      * @param DataObject $object The object to add to the ChangeSet
      * @param array|int $data Post data for this campaign form, or the ID of the campaign to add to
-     * @return HTTPResponse
      * @throws ValidationException
      */
-    public function addToCampaign($object, $data)
+    public function addToCampaign($object, $data): HTTPResponse
     {
         // Extract $campaignID from $data
         $campaignID = $this->getOrCreateCampaign($data);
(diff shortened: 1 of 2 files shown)


src/AddToCampaignHandler.php
```diff
@@ -42,7 +42,8 @@
  *
  * To use, you'd add an action to your LeftAndMain subclass, like this:
  *
- *     function addtocampaign($data, $form) {
+ *     function addtocampaign(array $data, Form $form): HTTPResponse
+ *     {
  *         $handler = AddToCampaignHandler::create($form, $data);
  *         return $handler->handle();
```
GuySartorelli: The `handle()` method in this class doesn't always return `HTTPResponse`

src/CampaignAdmin.php
```diff
@@ -730,13 +730,9 @@ public function getCampaignCreateForm()
...
     /**
-     * Save  handler
-     *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
+     * Save handler
      */
-    public function save($data, $form)
+    public function save(array $data, Form $form): HTTPResponse
```
GuySartorelli, line 735: `Security::PermissionFailure()` sometimes returns an array

GuySartorelli (changes requested):

src/AddToCampaignHandler.php
emteknetnz: Reverted

src/CampaignAdmin.php
emteknetnz: As noted in other PRs, it doesn't

GuySartorelli (approved):
Not to be merged until the other PRs are ready to avoid breaking CI

PR #246, opened 2022-10-18
API Update deprecations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10531


GuySartorelli (changes requested):
The version number is incorrect - use `1` major versioning instead. Also apply to any other deprecation notices in this module.

GuySartorelli (approved):

PR #258, opened 2023-03-13
FIX Replaced deprecated CSS and React functions

sabina-talipova:
### Description
- Replaced deprecated React function `componentWillReceiveProps` with `componentDidUpdate`;
- Replaced deprecated React function `componentWillMount` with `componentDidMount`;
- Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.
   -  Replaced divider `/` to `calc(*/*)` function

### Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1416


client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -59,20 +59,20 @@ class CampaignAdmin extends Component {
     this.handleToggleMessage = this.handleToggleMessage.bind(this);
   }
```
maxime-rainville, line 61: Tested this and it seems to be working fine.

maxime-rainville (approved):
LGTM.

PR #271, opened 2023-06-28
MNT ESLint issues

sabina-talipova:
## Description
Turn on some of the ESLint.
## Parent issue
-  https://github.com/silverstripe/silverstripe-admin/issues/1433


client/src/containers/CampaignAdmin/tests/CampaignAdmin-test.js
```diff
@@ -1,25 +1,27 @@
...
-
+import { Component as CampaignAdmin } from '../CampaignAdmin';
 
 let mockResponse;
 
 function makeProps(obj = {}) {
   return {
     sectionConfig: {
-      reactRoutePath: '',
+      reactRoutePath: '/campaigns',
```
emteknetnz, line 12: Seems wrong that we're changing this in a linting PR, this should be reverted. The tests were passing before so it's not a good sign if we need to change this now.

client/src/containers/CampaignAdmin/tests/CampaignAdmin-test.js
```diff
@@ -40,14 +42,6 @@ function makeProps(obj = {}) {
         view: null
       }
     },
-    sectionConfig: {
```
emteknetnz, line 43: We should not be removing stuff here in a linting PR, please revert

emteknetnz (changes requested):

client/src/containers/CampaignAdmin/tests/CampaignAdmin-test.js
sabina-talipova: I fix it to follow [no-dupe-keys](https://eslint.org/docs/latest/rules/no-dupe-keys) rule.
This object was initially created incorrectly and test shouldn't work properly. Object in JS cannot have duplicate keys, but if it has the last value will be saved. So this object after return looks like
```js
{
...
sectionConfig: {
   form: {
        campaignCreateForm: {
          schemaUrl: '/my/schema'
        }
      },
      reactRoutePath: '/campaigns',
    },
}
...
```
instead of 
```js
  sectionConfig: {
      reactRoutePath: '',
      reactRoutePath: '/campaigns',
      publishEndpoint: {
        url: '/',
        method: 'get',
      },
      form: {
        EditForm: {
          schemaUrl: 'schemaEditForm'
        },
        campaignCreateForm: {
          schemaUrl: '/my/schema'
        }
      }
    },
```
So I just combine all together. I don't know what we are going to achieve in this test logic, if you think that previous object was created correctly, I'll be glad to revert.

client/src/containers/CampaignAdmin/tests/CampaignAdmin-test.js
sabina-talipova: See comment https://github.com/silverstripe/silverstripe-campaign-admin/pull/271#discussion_r1253655654

emteknetnz (changes requested):
Please change the commit and PR prefix from FIX to MNT

emteknetnz (approved):

PR #286, opened 2023-10-24
MNT Behat test - removing implicitly added files

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-versioned/issues/422


sabina-talipova:
PR requires https://github.com/silverstripe/silverstripe-versioned/issues/422

maxime-rainville:
As discussed, we don't think there's a valid use case for removing item form campaign post publication.

I've created this follow up card instead: https://github.com/silverstripe/silverstripe-campaign-admin/issues/287

PR #346, opened 2025-03-04
ENH CMS 6 compatibility

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/895

Requirement is in framework so we don't need to include it here as well

It's also the CMS 5 version of the plugin so means that campaign-admin currently isn't installable with CMS 6


client/src/state/campaign/tests/CampaignActions-test.js
```diff
@@ -1,59 +0,0 @@
-/* global jest, jasmine, describe, it, expect, beforeEach */
-
-import configureMockStore from 'redux-mock-store';
```
emteknetnz, line 3: This test was removed because 'redux-mock-store' wasn't compatible with latest version of redux. There was a link in the failed test to a page where the maintainers or redux recommend testing with react-testing-library instead.  Since that kind of acts like a quasi behat test, and the fact we already have behat tests for campaign admin, I think we can safety just delete this test without a replacement

package.json
```diff
@@ -33,25 +33,20 @@
   "homepage": "https://github.com/silverstripe/silverstripe-campaign-admin",
   "dependencies": {
     "@popperjs/core": "^2.11.6",
-    "bootstrap": "^4.6.2",
+    "bootstrap": "^5.2.0",
     "classnames": "^2.3.2",
     "core-js": "^3.26.0",
     "deep-freeze-strict": "^1.1.1",
     "prop-types": "^15.8.1",
-    "react": "^18.2.0",
```
emteknetnz, line 41: Removed all of the `react*` deps because they're not needed as it'll automatically just use the deps from admin, and also because I ~could~ could not run `yarn build` with them here as they causes the 'Invalid hook call' issue.

client/src/legacy/AddToCampaignForm.js
```diff
@@ -0,0 +1,108 @@
+/* global window */
```
emteknetnz, line 1: This is JS that used to be in admin https://github.com/silverstripe/silverstripe-admin/pull/1857

_config/config.yml
```diff
@@ -12,3 +12,9 @@ SilverStripe\Admin\LeftAndMain:
 SilverStripe\CMS\Model\SiteTree:
   extensions:
     - SilverStripe\CampaignAdmin\SiteTreeExtension
+
+# Allow users only only have access to the campaign-admin, and not all CMS access
+# i.e. CMS_ACCESS_LeftAndMain, to still be able to manage changesets
+SilverStripe\Versioned\ChangeSet:
```
emteknetnz, line 14: These used to be done in Versioned https://github.com/silverstripe/silverstripe-versioned/pull/482

The plural_name is used [here](https://github.com/silverstripe/silverstripe-cms/pull/3062/files#diff-8de4622cc87cca016e9e2d317482ab37ac8a29a900134ad8ffad3d89ee108937R1575) with a call to ChangeSet::singleton()->i18n_pluralise()

We may as well also keep the singlular_name for consistency

composer.json
```diff
@@ -24,8 +24,7 @@
         "php": "^8.3",
         "silverstripe/admin": "^3",
         "silverstripe/framework": "^6",
-        "silverstripe/versioned": "^3",
-        "silverstripe/vendor-plugin": "^2"
```
emteknetnz, line 28: Using the vendor-plugin defined in framework instead

src/Extensions/CMSPageEditControllerExtension.php
```diff
@@ -0,0 +1,114 @@
+<?php
```
emteknetnz: This class contains a bunch of code that used to be in silverstripe/cms https://github.com/silverstripe/silverstripe-cms/pull/3027

tests/php/AddToCampaignValidatorTest.php
```diff
@@ -68,6 +68,8 @@ public function testValidationFailed()
                 'fieldName' => 'Campaign',
                 'messageType' => 'error',
                 'messageCast' => 'text',
+                'modelClass' => '',
```
emteknetnz, line 71: Validators in CMS 6 pass through extra info so that it can be seen in CLI

src/Extensions/AssetAdminExtension.php
```diff
@@ -0,0 +1,104 @@
+<?php
```
emteknetnz: This contains code from asset-admin https://github.com/silverstripe/silverstripe-asset-admin/pull/1512

src/Extensions/FileFormFactoryExtension.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
emteknetnz, line 1: This contains code from asset-admin https://github.com/silverstripe/silverstripe-asset-admin/pull/1512

src/Extensions/HideInCampaignsExtension.php
```diff
@@ -0,0 +1,13 @@
+<?php
```
emteknetnz, line 1: This was used in both assets - https://github.com/silverstripe/silverstripe-assets/pull/660

And elemental - https://github.com/silverstripe/silverstripe-elemental/pull/1283

package.json
GuySartorelli: I assume you mean "because I _couldn't_ run `yarn build` with them there"

src/Extensions/AddToCampaignExtension.php
```diff
@@ -0,0 +1,130 @@
...
+
+    protected function updateClientConfig(array &$clientConfig): void
+    {
+        $clientConfig = ArrayLib::array_merge_recursive($clientConfig, [
+            'form' => [
+                'addToCampaignForm' => [
+                    'schemaUrl' => $this->getOwner()->Link('schema/addToCampaignForm')
+                ],
+            ],
+        ]);
```
GuySartorelli:
```suggestion
        $clientConfig['form']['addToCampaignForm']['schemaUrl'] = $this->getOwner()->Link('schema/addToCampaignForm');
```

_config/config.yml
```diff
@@ -9,6 +9,10 @@ SilverStripe\Admin\LeftAndMain:
...
-SilverStripe\CMS\Model\SiteTree:
-  extensions:
-    - SilverStripe\CampaignAdmin\SiteTreeExtension
+# Allow users only only have access to the campaign-admin, and not all CMS access
+# i.e. CMS_ACCESS_LeftAndMain, to still be able to manage changesets
+SilverStripe\Versioned\ChangeSet:
+  required_permission:
+    - CMS_ACCESS_CampaignAdmin
+  singular_name: 'Campaign'
+  plural_name: 'Campaigns'
```
GuySartorelli: I removed the singular and plural name changes and didn't see anything change. What is this needed for?

client/src/legacy/AddToCampaignForm.js
```diff
@@ -0,0 +1,108 @@
...
+    _renderModal(isOpen) {
+      const handleHide = () => this.close();
+      const handleSubmit = (...args) => this._handleSubmitModal(...args);
+      const id = $('form.cms-edit-form :input[name=ID]').val();
+      const sectionConfigKey = 'SilverStripe\\CMS\\Controllers\\CMSPageEditController';
+      const store = window.ss.store;
+      const sectionConfig = store.getState().config.sections
+        .find((section) => section.name === sectionConfigKey);
+      const modalSchemaUrl = `${sectionConfig.form.addToCampaignForm.schemaUrl}/${id}`;
+      const title = i18n._t('Admin.ADD_TO_CAMPAIGN', 'Add to campaign');
```
GuySartorelli: This and any other i18n keys in this PR need to be updated to reflect that they're in campaign admin now, not in those original modules.

client/src/state/campaign/tests/CampaignActions-test.js
GuySartorelli, line 1: Why is this being removed?

_config/config.yml
GuySartorelli: https://github.com/silverstripe/silverstripe-campaign-admin/pull/346#discussion_r1980653297 links to a closed PR

src/Extensions/AddToCampaignExtension.php
```diff
@@ -0,0 +1,130 @@
...
+{
+    private static array $allowed_actions = [
+        'addToCampaignForm',
+    ];
+
+    /**
+     * Data class for managed data objects (e.g. pages, files) that are managed by the campaign admin.
+     * This must be set via config elsewhere i.e. yml config or else a RuntimeException will be thrown
+     */
+    private static ?string $campaign_admin_managed_data_class = null;
```
GuySartorelli: Can you please explain why setting it here doesn't work? That sounds a bit weird, almost like there's some underlying bug in the config system or something, but I assume it's much more mundane than that.

GuySartorelli: I also can only see where it's set when `asset-admin` and `cms` are installed, implying that there will always be a runtime exception if those modules aren't installed?

src/Extensions/AddToCampaignExtension.php
```diff
@@ -0,0 +1,130 @@
...
+use SilverStripe\Core\Extension;
+use SilverStripe\Core\Validation\ValidationResult;
+use SilverStripe\Forms\Form;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * Extension applied to a Controller to add the ability to add items to a campaign.
+ * This extension will only be applied if the `campaign-admin` module is installed.
+ *
+ * @extends Extension<AssetAdmin>
```
GuySartorelli:
```suggestion
 * @extends Extension<AssetAdmin|CMSPageEditController>
```
Or could it theoretically be added to any `LeftAndMain`?

src/Extensions/FileFormFactoryExtension.php
```diff
@@ -0,0 +1,39 @@
...
+
+use SilverStripe\AssetAdmin\Forms\FileFormFactory;
+use SilverStripe\Assets\File;
+use SilverStripe\Core\Extension;
+use SilverStripe\Forms\FormAction;
+use SilverStripe\Security\Permission;
+
+/**
+ * Extension that updates the Popover menu of `FileFormFactory`.
+ * This extension will only be applied if the `campaign-admin` module is installed.
```
GuySartorelli:
```suggestion
```
And remove similar wording from other copied code, if it's there

GuySartorelli (changes requested):
Tested locally - functionality all seems to be working as expected.

GuySartorelli (changes requested):
Just noticed the `draft` label displayed next to records in the campaigns doesn't have the correct styling - it's likely not using the new status badge API to get the markup for those.

src/Extensions/AddToCampaignExtension.php
emteknetnz: Updated

_config/config.yml
emteknetnz: Have removed, turns out there are translations in Versioned which will call Changesets "Campaigns", so we don't need the config

client/src/legacy/AddToCampaignForm.js
emteknetnz: Updated - as far as I can tell this was the only key that needed to be updated

client/src/state/campaign/tests/CampaignActions-test.js
emteknetnz: There's a comment explaining why here https://github.com/silverstripe/silverstripe-campaign-admin/pull/346/files#r1980615400

src/Extensions/AddToCampaignExtension.php
emteknetnz: Updated to LeftAndMain

src/Extensions/AddToCampaignExtension.php
emteknetnz: Have updated the comment to clarify

src/Extensions/FileFormFactoryExtension.php
emteknetnz: Removed. Also removed from AddToCampaignExtension which had the same comment

src/Extensions/AddToCampaignExtension.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Get the managed data class for this extension
+     */
+    private function getManagedDataClass(): string
+    {
+        $owner = $this->getOwner();
+        $property = 'campaign_admin_managed_data_class';
+        $dataClass = $owner->config()->get($property);
+        if (!is_a($dataClass, DataObject::class, true)) {
+            throw new RuntimeException("$owner.$property is not a DataObject subclass.");
```
GuySartorelli: `$owner` here should not be trusted to evaluate to a class name. Use `get_class($owner)`

GuySartorelli (changes requested):
Just the one change - sorry, should have noticed it last time

src/Extensions/AddToCampaignExtension.php
GuySartorelli: Also, if it needs to explicitly be a subclass (i.e. can't be `DataObject` itself), use `is_subclass_of()` - but if `DataObject` is allowed, update the message to "is not DataObject or a subclass" or similar.

GuySartorelli:
Also please respond to https://github.com/silverstripe/silverstripe-campaign-admin/pull/346#pullrequestreview-2672285907

src/Extensions/AddToCampaignExtension.php
emteknetnz: Updated

emteknetnz:
> Just noticed the draft label displayed next to records in the campaigns doesn't have the correct styling - it's likely not using the new status badge API to get the markup for those.

Have updated styling

GuySartorelli (approved):
LGTM

PR #360, opened 2025-07-30
ENH Reduce queries by using new *byList methods

GuySartorelli:
Some queries can be combined using the new filterByList and excludeByList queries on DataList. This reduces latency overhead and provides an opportunity for the database optimiser to do its thing

> [!IMPORTANT]
> Requires https://github.com/silverstripe/silverstripe-framework/pull/11804

## Issue

- https://github.com/silverstripe/.github/issues/416

### DIFF
--- a/src/Extensions/CMSMainExtension.php
+++ b/src/Extensions/CMSMainExtension.php
@@ -11,14 +11,15 @@ class CMSMainExtension extends Extension
 {
     function updateArchiveWarningMessage(string &$message, array $descendants)
     {
-        $inChangeSetIDs = ChangeSetItem::get()->filter([
+        $inChangeSetList = ChangeSetItem::get()->filter([
             'ObjectID' => $descendants,
             'ObjectClass' => SiteTree::class
-        ])->column('ChangeSetID');
+        ]);
         $affectedChangeSetCount = 0;
-        if (count($inChangeSetIDs ?? []) > 0) {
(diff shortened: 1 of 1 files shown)


emteknetnz (approved):

src/Extensions/CMSMainExtension.php
```diff
@@ -11,14 +11,15 @@ class CMSMainExtension extends Extension
...
     function updateArchiveWarningMessage(string &$message, array $descendants)
     {
-        $inChangeSetIDs = ChangeSetItem::get()->filter([
+        $inChangeSetList = ChangeSetItem::get()->filter([
             'ObjectID' => $descendants,
             'ObjectClass' => SiteTree::class
-        ])->column('ChangeSetID');
+        ]);
         $affectedChangeSetCount = 0;
         if (count($inChangeSetIDs ?? []) > 0) {
```
emteknetnz: This is still using the old variable, this condition will always be false

emteknetnz (changes requested):

src/Extensions/CMSMainExtension.php
GuySartorelli: Good catch, thank you. Updated to use `exists()` which will be a little more efficient than count and was probably what I originally intended to do

emteknetnz (approved):

PR #367, opened 2025-08-18
FIX Generate correct archive message

GuySartorelli:
CMSMain doesn't include the record in the descendants array anymore, but we do need to know if the current record has a campaign associated with it or not.

Reflects a change in https://github.com/silverstripe/silverstripe-cms/pull/3104

> [!IMPORTANT]
> Tests for this will fail without updating the dependency constraint to the patch for the above, or it will fail on prefer-lowest CI builds. 
> I'd like to add a test for this - but I can't do that reliably until the above PR is merged and tagged.

## Issue

- https://github.com/silverstripe/silverstripe-cms/issues/3103

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,9 @@
         "silverstripe/standards": "^1",
         "phpstan/extension-installer": "^1.3"
     },
+    "conflict": {
+        "silverstripe/cms": "<6.0.2"
+    },
     "extra": {
         "expose": [
             "client/dist",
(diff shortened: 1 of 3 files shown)


emteknetnz (changes requested):
> I'd like to add a test for this - but I can't do that reliably until the above PR is merged and tagged.

I've merged the corresponding CMS PR

composer.json
```diff
@@ -34,6 +34,9 @@
         "silverstripe/standards": "^1",
         "phpstan/extension-installer": "^1.3"
     },
+    "conflict": {
+        "silverstripe/cms": "<6.0.2"
+    },
```
GuySartorelli, line 39: https://github.com/silverstripe/silverstripe-cms/releases/tag/6.0.2

emteknetnz (approved):
Works well

Issue #366, opened 2025-08-11
Accessibility: Cannot open or close "help" information


GuySartorelli:
In campaign admin, there's a `?` button with the aria label "Show help" which when clicked displays additional information. When the information is displayed, the `?` button is replaced with a `x` with aria label "Hide help".

Both of these buttons _are_ in the tab focus order, but:
1. There is no visible focus state
2. When they have focus, neither the enter nor space keys do anything
3. They're both `<a>` tags but should be `<button>`.

## Acceptance criteria

- [ ] Appropriate `:focus-visible` CSS is applied (see https://github.com/silverstripe/.github/issues/443)
- [x] Pressing `space` and `enter` keys is the same as clicking the button.
- [x] `<a>` tag is replaced with `<button>` tag
- [x] The changelog clearly calls out any changes to markup selectors so people can adjust their custom css/js as necessary

## PRs
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/371
- https://github.com/silverstripe/developer-docs/pull/819

GuySartorelli:
PRs merged

PR #371, opened 2025-09-02
ENH Improve help button accessibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-campaign-admin/issues/366


GuySartorelli:
@emteknetnz Don't forget to set as "ready for review'

GuySartorelli:
After activating the `?` button with the keyboard, I have to tab _backwards_ to actually get any of that help information.

Instead, it would make sense for my focus to jump to the close button, so the user can then tab through the help information. If you do this, probably the `?` button shouldn't exist when the help information is displayed (one button to open, one button to close). Otherwise a user may see the button and go "oh I should click that for more information" only for the information to actually be hidden (especially since the aria label is "show help").

The "show help" (`?`) button should probably also be before anything else so it's in the same place in the tab order as the close help button.

Please also create a changelog PR as per the acceptance criteria

GuySartorelli:
> The "show help" (`?`) button should probably also be before anything else so it's in the same place in the tab order as the close help button.

Looks like this bit isn't done? Was that intentionally omitted or just missed off?

With this PR the "close help" button is _before_ "add campaign" in the tab order (which is correct) - but the "open help" button is _after_ "add campaign" in the tab order (which seems wrong). This means if I close help (which then correctly takes my focus to the open help button) and then hit tab, it looks like there's nothing there - I won't ever get to the "add campaign" button.

emteknetnz:
> Was that intentionally omitted or just missed off?

Intentionally omitted, sorry forgot to put in comment. I think the tab order is correct as top->down/left->right is the natural tab order, and the help button is to the right of the Add Campaign button.

The Add Campaign button should also be the first thing tabbed when normally using the page (most of the time people won't be using the the help button since they've used the campaign admin plenty of times before) to since it's the primary action, whereas the help button is there to support it.

GuySartorelli:
> Intentionally omitted, sorry forgot to put in comment. I think the tab order is correct as top->down/left->right is the natural tab order, and the help button is to the right of the Add Campaign button.

I recommend moving the help button.

> The Add Campaign button should also be the first thing tabbed when normally using the page (most of the time people won't be using the the help button since they've used the campaign admin plenty of times before) to since it's the primary action, whereas the help button is there to support it.

1. The help is displayed by default
2. Dismissing the help the requires you to go _backwards_ in the tab order to get to any of the campaign functionality (which requires you to know it's there, which blind users may not). This is the bad thing and is a whole accessibility problem on its own. In the worst case users won't even know they have to go backwards to get to that content, so it will appear as though that content isn't there or can't be accessed.

Because of this, I suggest that one of these happens:
1. Move the "show help" button to be first, which puts it in the same place in the tab order as the "close help" button which means users won't be skipping over content by closing the help info, OR
2. Move the help info and "close help" button to be after everything else (though this will be weird for sighted users)

emteknetnz:
> Dismissing the help the requires you to go backwards in the tab order to get to any of the campaign functionality

That's incorrect? When you go to `/admin/campaigns` your focus start in the top-left of the cms i.e. the first time you press tab you'll focus on the name of the website "Your site name".  As you tab through things you'll go through the left-hand nav, to the main content panel.  

In the content panel the tab order is the close button first, the help panel links, then the add to campaign button.  You don't tab backwards to get to the close button.

GuySartorelli:
>> Dismissing the help the requires you to go backwards in the tab order to get to any of the campaign functionality
>
> That's incorrect? When you go to `/admin/campaigns` your focus start in the top-left of the cms i.e. the first time you press tab you'll focus on the name of the website "Your site name".  As you tab through things you'll go through the left-hand nav, to the main content panel.  

For brevity I was skipping describing all the irrelevant stuff that happens before you get to the content panel.

> In the content panel the tab order is the close button first, the help panel links, then the add to campaign button.  You don't tab backwards to get to the close button.

When you close the help information, your focus is then on the "open help" button, which is _after_ the add to campaign button. So if you dismiss the help information, your focus has skipped over the add to campaign button. To get to it, you'll have to use `shift+tab` (go backwards)

emteknetnz:
Ok that makes sense, have updated

GuySartorelli (approved):
Seems good. Works locally.

Issue #365, opened 2025-08-11
Usability: Add ability to search/filter by campaigns


GuySartorelli:
The campaign admin currently has no search or filter functionality.

This is especially a pain given all published campaigns are displayed alongside the unpublished ones.

## Notes

- Ideally this uses `searchable_fields` and the regular `SearchContext` functionality.

## Acceptance criteria

- [x]  A search/filter functionality (similar to everywhere else in the CMS, asset admin is a good example) is implemented
- [x] You can search/filter by title, status, and published date (with a between filter)

## PRs

[**Kitchen sink CI run**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/19150055935)

- https://github.com/silverstripe/silverstripe-versioned/pull/533
- https://github.com/silverstripe/silverstripe-admin/pull/2054
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/380
- https://github.com/silverstripe/developer-docs/pull/844

emteknetnz:
This seems out of scope for accessibility - should be moved to the regular backlog?

GuySartorelli:
Scope is up to the product owner. This came out of the accessibility testing the testing team did.

PR #380, opened 2025-11-06
NEW Add search/filter to campaign admin

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-versioned/pull/533 and https://github.com/silverstripe/silverstripe-admin/pull/2054 for behat to go green

image

The datetime field stretching too far is [a known issue](https://github.com/silverstripe/silverstripe-admin/issues/1997)

## Issue

- https://github.com/silverstripe/silverstripe-campaign-admin/issues/365


client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -51,6 +55,11 @@ class CampaignAdmin extends Component {
       },
     });
 
+    this.searchCampaignsApi = backend.createEndpointFetcher({
+      ...props.sectionConfig.searchCampaignsEndpoint,
+      payloadSchema: {},
+    });
```
GuySartorelli, line 61: Similar to other "api" endpoints above.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -87,6 +99,30 @@ class CampaignAdmin extends Component {
...
+    this.setState({ filters });
+    // If there are no filters, or the filter values are empty, just fetch everything
+    if (!hasFilters(filters) || Object.values(filters).filter((val) => val || val === 0).length === 0) {
+      return this.fetchCampaignsList();
+    }
+    return this.props.campaignActions.searchCampaigns(
+      this.props.sectionConfig.treeClass,
+      this.searchCampaignsApi,
+      filters
+    );
```
GuySartorelli, line 123: Using similar mechanism to other actions already in this component. Doing this means we can hook into existing state stuff to re-render the campaigns list with the results.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -87,6 +99,30 @@ class CampaignAdmin extends Component {
...
+    this.setState({ showSearch: false });
+    this.handleDoSearch({});
+  }
+
+  handleDoSearch(filters) {
+    this.setState({ filters });
+    // If there are no filters, or the filter values are empty, just fetch everything
+    if (!hasFilters(filters) || Object.values(filters).filter((val) => val || val === 0).length === 0) {
+      return this.fetchCampaignsList();
+    }
```
GuySartorelli, line 118: Clearing search filters.

Note that removing a single search filter by clicking the `x` for that filter (example screenshot below) sets the filter to `undefined` in this filter object.

If the only filters are `undefined` and we try to use the search campaigns action below, we get an error. The correct behaviour in that scenario is the same as clearing filters anyway - we have no filters so we should get the full list.

image

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -433,11 +469,27 @@ By removing this item all linked items will be removed unless used elsewhere.`;
       identifier: 'Campaign.IndexView',
     };
 
+    const showSearch = hasFilters(this.state.filters) || this.state.showSearch;
```
GuySartorelli, line 472: Same logic asset admin uses

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -433,11 +469,27 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
+
     return (
       <div className="fill-height" aria-expanded="true">
         <Toolbar>
           <BreadcrumbComponent multiline />
+          <div className="campaign--toolbar__extra pull-xs-right fill-width vertical-align-items">
+            <SearchToggle toggled={showSearch} onToggle={this.toggleSearch} />
+          </div>
```
GuySartorelli, line 480: Putting it in the toolbar to match asset admin.
Putting it just above the gridfield-like-thing when I tried it didn't look great, required a bunch of additional CSS, and isn't necessary since we only ever have one list that can be filtered here anyway.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -433,11 +469,27 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
+          onSearch={this.handleDoSearch}
+          id="CampaignSearchForm"
+          formSchemaUrl={sectionConfig.form.campaignSearchForm.schemaUrl}
+          onHide={this.handleClearSearch}
+          displayBehavior="HIDEABLE"
+          filters={this.state.filters}
+          filterPrefix="Search__"
+          addFilterPrefix
+          name={sectionConfig.searchCampaignsGeneralField}
+        />}
```
GuySartorelli, line 492: Combination of asset admin and cmsmain/gridfield props - see description in https://github.com/silverstripe/silverstripe-admin/pull/2054

client/src/containers/CampaignAdmin/CampaignAdmin.scss
```diff
@@ -173,3 +173,11 @@
...
   flex-direction: row-reverse;
 }
+
+.campaign--toolbar__extra {
+  width: auto;
+
+  button:last-of-type {
+    margin-right: -10px;
+  }
+}
```
GuySartorelli, line 183: Same CSS is used in asset admin on a class called `asset-admin__toolbar-extra` and is needed to pull the search button to the right of the toolbar

client/src/state/campaign/CampaignActions.js
```diff
@@ -129,3 +129,27 @@ export function removeCampaignItem(removeItemApi, campaignId, itemId) {
...
+      })
+      .catch((error) => {
+        dispatch({
+          type: RECORD_ACTION_TYPES.FETCH_RECORDS_FAILURE,
+          payload: { error, recordType },
+        });
+        throw error;
+      });
+  };
+}
```
GuySartorelli, line 155: The general structure and dispatch types of this matches `fetchRecords()` in https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/state/records/RecordsActions.js which is what's used to fetch the records initially.

Reusing the same dispatch types means the list of campaigns is re-rendered using the data we fetch from the search endpoint.

src/CampaignAdmin.php
```diff
@@ -38,6 +41,8 @@ class CampaignAdmin extends LeftAndMain implements PermissionProvider
         'EditForm',
         'campaignEditForm',
         'campaignCreateForm',
+        'campaignSearchForm',
```
GuySartorelli, line 44: We don't use `campaignSearchForm` directly - but we need it to be in `allowed_actions` so that `$this->Link('schema/campaignSearchForm')` doesn't complain at us.

src/CampaignAdmin.php
```diff
@@ -720,6 +744,34 @@ public function getCampaignCreateForm()
...
+    public function searchCampaigns(HTTPRequest $request): HTTPResponse
+    {
+        $searchContext = ChangeSet::singleton()->getDefaultSearchContext();
+        $filters = $request->requestVar('filters');
+        $form = $this->getCampaignSearchForm($searchContext);
+        $filterArguments = $form->prepareValuesForSearchContext($filters);
+        $results = $searchContext->getQuery($filterArguments, existingQuery: $this->getRawListItems());
+        $results = $results->filterByCallback(function ($item) {
+            /** @var ChangeSet $item */
+            return ($item->canView());
```
emteknetnz:
```suggestion
            return $item->canView();
```

src/CampaignAdmin.php
```diff
@@ -720,6 +744,34 @@ public function getCampaignCreateForm()
...
+    public function getCampaignSearchForm(?SearchContext $searchContext = null): SearchContextForm
+    {
+        if (!$searchContext) {
+            $searchContext = ChangeSet::singleton()->getDefaultSearchContext();
+        }
+        $form = SearchContextForm::create($this, $searchContext, 'campaignSearchForm');
+        return $form;
+    }
+
+    public function searchCampaigns(HTTPRequest $request): HTTPResponse
```
emteknetnz, line 762: Add a quick doc block description of what this method does

src/CampaignAdmin.php
```diff
@@ -720,6 +744,34 @@ public function getCampaignCreateForm()
         return $form;
     }
 
+    public function getCampaignSearchForm(?SearchContext $searchContext = null): SearchContextForm
```
emteknetnz, line 750: Add a quick doc block description of what this method does

tests/behat/features/filter-campaigns.feature
```diff
@@ -0,0 +1,69 @@
...
+  I want to search/filter campaigns within the CMS
+  So that I can find specific campaigns quickly
+
+  Background:
+    Given a "ChangeSet" "First Campaign" with "Description"="this is a campaign"
+      And a "ChangeSet" "Second Campaign" with "Description"="lorem ipsum" and "State"="published" and "PublishDate"="2025-01-01 12:00:00"
+      And a "ChangeSet" "Third Campaign" with "Description"="lorem ipsum" and "State"="published" and "PublishDate"="2025-05-06 12:00:00"
+      And the "group" "CAMPAIGNS_EDITOR" has permissions "Access to 'Campaigns' section"
+      And I am logged in as a member of "CAMPAIGNS_EDITOR" group
+      And I go to "/admin/campaigns"
```
emteknetnz, line 13:
```suggestion
    Given a "ChangeSet" "First Campaign" with "Description"="this is a campaign"
    And a "ChangeSet" "Second Campaign" with "Description"="lorem ipsum" and "State"="published" and "PublishDate"="2025-01-01 12:00:00"
    And a "ChangeSet" "Third Campaign" with "Description"="lorem ipsum" and "State"="published" and "PublishDate"="2025-05-06 12:00:00"
    And the "group" "CAMPAIGNS_EDITOR" has permissions "Access to 'Campaigns' section"
    And I am logged in as a member of "CAMPAIGNS_EDITOR" group
    And I go to "/admin/campaigns"
```

And unident all the other bits in this file

emteknetnz (changes requested):

src/CampaignAdmin.php
GuySartorelli: Done

src/CampaignAdmin.php
GuySartorelli: Done

src/CampaignAdmin.php
GuySartorelli: Done

tests/behat/features/filter-campaigns.feature
GuySartorelli: Indentation is all over the place in feature files. Unless we're going to adopt a specific style guide for them I don't think it's worth either of our time worrying about it.

Even within this module (and even within some individual feature files) there's already a mix of indented and not-indented `And` lines (though it looks like the slight majority in this module _do_ indent them).

emteknetnz (approved):

Issue #381, opened 2025-11-10
Convert campaign-admin to functional components


emteknetnz:
There are several react class components that needs to be converted to react functional components)

- [ ] Components are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## PRs
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/382

GuySartorelli:
PR merged

PR #382, opened 2025-11-10
ENH Refactor campaign-admin from class to functional component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-campaign-admin/issues/381


client/src/components/IntroScreen/IntroScreen.js
```diff
@@ -1,39 +1,28 @@
...
+  onClose = noop,
+  focusCloseButton = false,
+}) => {
+  const closeButtonRef = useRef(null);
 
-    this.handleClose = this.handleClose.bind(this);
-    this.closeButtonRef = React.createRef();
-  }
-
-  componentDidMount() {
```
emteknetnz, line 16: Removed as its duplicated in componentDidUpdate

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -234,127 +238,26 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
-  /**
-   * Hook to allow customisation of components being constructed
-   * by the Campaign DetailEdit FormBuilderLoader.
-   *
-   * @param {object} Custom Component constructor.
-   * @param {object} props Props passed from FormBuilderLoader.
-   *
-   * @return {object} Instantiated React component
-   */
-  campaignEditCreateFn(Custom, props) {
```
emteknetnz, line 312: Was unused so removed

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -234,127 +238,26 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
-  /**
-   * Hook to allow customisation of components being constructed
-   * by the Campaign creation FormBuilderLoader.
-   *
-   * @param {object} Custom Component constructor.
-   * @param {object} props Props passed from FormBuilderLoader.
-   *
-   * @return {object} Instantiated React component
-   */
-  campaignAddCreateFn(Custom, props) {
```
emteknetnz, line 340: Was unused so removed

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -24,110 +24,68 @@ import CampaignAdminList from './CampaignAdminList';
...
-    );
-  }
+  /**
+   * Generate route with the given id and view
+   *
+   * @param {number} id
+   * @param {string} view
+   * @return {string}
+   */
+  const getActionRoute = (id, view) => {
```
emteknetnz, line 80: Moved up as it's used by some other functions

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -24,110 +24,68 @@ import CampaignAdminList from './CampaignAdminList';
...
+  };
 
-  setBreadcrumbs(view, id, title) {
-    const { sectionConfig: { reactRoutePath } } = this.props;
+  const setBreadcrumbs = (view, id, breadcrumbTitle) => {
+    const { reactRoutePath } = sectionConfig;
 
     // Set root breadcrumb
-    const breadcrumbs = [{
+    const newBreadcrumbs = [{
```
emteknetnz, line 89: renamed as there's a breadcrumbs prop

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -24,110 +24,68 @@ import CampaignAdminList from './CampaignAdminList';
...
+   * @return {string}
+   */
+  const getActionRoute = (id, view) => {
+    const { reactRoutePath } = sectionConfig;
+    return joinUrlPaths('/', reactRoutePath, `/set/${id}/${view}`);
+  };
 
-  setBreadcrumbs(view, id, title) {
-    const { sectionConfig: { reactRoutePath } } = this.props;
+  const setBreadcrumbs = (view, id, breadcrumbTitle) => {
```
emteknetnz, line 85: renamed param as there's a title prop

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -188,41 +187,46 @@ class CampaignAdmin extends Component {
...
+  const handleCreateCampaignSubmit = (data, action, submitFn) => {
     const promise = submitFn();
     if (!promise) {
       throw new Error('Promise was not returned for submitting');
     }
     return promise
       .then((response) => {
-        const hasErrors = this.hasErrors(response);
-        if (action === 'action_save' && !hasErrors) {
+        const hasErrorsResult = hasErrors(response);
```
emteknetnz, line 195: Renamed var as there's a hasErrors method

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -136,49 +94,90 @@ class CampaignAdmin extends Component {
...
-   *
-   * @param {number} id
-   * @param {string} view
-   * @return {string}
-   */
-  getActionRoute(id, view) {
-    const { reactRoutePath } = this.props.sectionConfig;
-    return joinUrlPaths('/', reactRoutePath, `/set/${id}/${view}`);
-  }
+  const fetchCampaignsList = () => {
```
emteknetnz, line 116: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -136,49 +94,90 @@ class CampaignAdmin extends Component {
...
+    const endpoint = sectionConfig.readCampaignsEndpoint;
+    const fetchURL = endpoint.url;
+    return recordActionsObj.fetchRecords(
+      sectionConfig.treeClass,
+      endpoint.method,
+      fetchURL
+    );
+  };
+
+  const handleDoSearch = (filtersParam) => {
```
emteknetnz, line 126: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -136,49 +94,90 @@ class CampaignAdmin extends Component {
...
+    return campaignActionsObj.searchCampaigns(
+      sectionConfig.treeClass,
+      searchCampaignsApi,
+      filtersParam
+    );
+  };
 
-  handleBackButtonClick(event) {
-    const { breadcrumbs, router: { navigate } } = this.props;
+  const toggleSearch = () => {
```
emteknetnz, line 139: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -136,49 +94,90 @@ class CampaignAdmin extends Component {
...
+    setShowSearch(prevShowSearch => !prevShowSearch);
+  };
 
+  const handleClearSearch = () => {
```
emteknetnz, line 143: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -444,126 +379,145 @@ class CampaignAdminList extends Component {
...
+        </AccordionBlock>
       </div>
     );
-  }
-}
+  });
+
+  const body = <Accordion>{accordionBlocks}</Accordion>;
+
+  const loadingSpinner = !loading && [
```
emteknetnz: Renamed from loading to loadingSpinner as it was a poorly named variable

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -91,29 +56,80 @@ class CampaignAdminList extends Component {
...
-      href: joinUrlPaths('/', reactRoutePath, `set/${campaignId}/show`),
+      href: joinUrlPaths('/', sectionConfig.reactRoutePath, `set/${campaignId}/show`),
     });
 
-    actions.setBreadcrumbs(breadcrumbs);
-  }
+    breadcrumbsActions.setBreadcrumbs(breadcrumbs);
+  };
+
+  const getItems = () => {
```
emteknetnz, line 69: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -91,29 +56,80 @@ class CampaignAdminList extends Component {
...
+    if (record && record._embedded) {
+      return record._embedded.items;
+    }
+    return null;
+  };
 
-  getSelectedItem() {
-    const { campaign: { changeSetItemId } } = this.props;
-    const items = this.getItems() || [];
+  const getPlaceholderGroups = () => {
```
emteknetnz, line 76: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -91,29 +56,80 @@ class CampaignAdminList extends Component {
...
+
+    return groups;
+  };
+
+  /**
+   * Group items for changeset display
+   *
+   * @return {object}
+   */
+  const groupItemsForSet = () => {
```
emteknetnz, line 94: Moved up above other things that call it

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -124,14 +140,19 @@ class CampaignAdminList extends Component {
     }
 
     return selected;
-  }
+  };
+
+  const handleRemoveItem = () => {
```
emteknetnz, line 145: Moved up above other things that call it

client/src/components/IntroScreen/IntroScreen.js
```diff
@@ -1,39 +1,28 @@
...
-      !prevProps.focusCloseButton &&
-      this.closeButtonRef.current
-    ) {
-      this.closeButtonRef.current.focus();
+  useEffect(() => {
+    if (focusCloseButton && closeButtonRef.current) {
+      closeButtonRef.current.focus();
     }
-  }
+  }, [focusCloseButton]);
```
GuySartorelli, line 19: Does `closeButtonRef.current` need to be in the dependencies array as well?

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -631,8 +548,6 @@ function mapStateToProps(state, ownProps) {
   return {
     previewState: viewMode.activeState,
     config: state.config,
-    campaignId: state.campaign.campaignId,
-    view: state.campaign.view,
```
GuySartorelli, line 635: These are missing. Looking through the original code it looks like they weren't used so removing them is correct, but please point out changes like this for future refactors.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -387,114 +290,145 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
+  };
+
+  const prevPropsRef = useRef();
+
+  useEffect(() => {
+    // Ensure default breadcrumbs are setup
+    if (breadcrumbs.length === 0) {
+      setBreadcrumbs(router.params.view, router.params.id, title);
+    }
+  }, [breadcrumbs, router.params.view, router.params.id, title, setBreadcrumbs]);
```
GuySartorelli:
```suggestion
  useEffect(() => {
    // Ensure default breadcrumbs are setup
    if (breadcrumbs.length === 0) {
      setBreadcrumbs(router.params.view, router.params.id, title);
    }
  }, []);
```

Looks like this is meant to replace `componentDidMount`, but it has a bunch of dependencies? Replacing `componentDidMount()` usually has just an empty array instead. Am I missing some context here?

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -387,114 +290,145 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
-  addCampaign() {
-    const path = this.getActionRoute(0, 'create');
-    this.props.router.navigate(path);
-  }
+  const addCampaign = () => {
+    const path = getActionRoute(0, 'create');
+    router.navigate(path);
+  };
+
+  const prevPropsRef = useRef();
```
GuySartorelli: The only ref before was the help button one. I can see this is used in the `useEffect` that replaces `componentDidUpdate()` which means refactoring was probably done. Can you please explain the changes and why a new `if (prevPropsRef.current) {` condition was added to that?

Kinda feels like it's being used instead of a dependencies array for the use effect hook - except those dependencies are there too.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -387,114 +290,145 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
+      );
+      if (hasChangedRoute) {
+        setBreadcrumbs(router.params.view, router.params.id, title);
+      }
+    }
+
+    // Focus help button when showMessage toggles from true to false
+    if (prevPropsRef.current && prevPropsRef.current.showMessage && !showMessage && helpButtonRef.current) {
+      helpButtonRef.current.focus();
+    }
```
GuySartorelli: This condition has changed. Use to be:

```js
if (prevProps.showMessage && !this.props.showMessage && this.helpButtonRef.current) {
```

now:

```js
if (prevPropsRef.current && prevPropsRef.current.showMessage && !showMessage && helpButtonRef.current) {
```

Ignoring the `prevPropsRef` which is discussed in a different comment, `!showMessage` was added when it wasn't there before.

client/src/containers/CampaignAdmin/CampaignAdmin.js
GuySartorelli: Also `filters` renamed - presumably to avoid collision with the state with the same name.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -136,49 +95,90 @@ class CampaignAdmin extends Component {
...
+   * @param {object} response
+   * @returns {boolean}
+   */
+  const hasErrors = (response) => {
+    if (response.errors && response.errors.length) {
+      return true;
+    }
+    const responseState = response.state;
+
+    if (!responseState) {
```
GuySartorelli, line 168: `responseState` renamed from `state` - though it doesn't materially matter so leave as is. Just took me a second going "wait we weren't checking against this variable before" lol

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -188,41 +188,46 @@ class CampaignAdmin extends Component {
...
       event.preventDefault();
     }
-  }
+  };
+
+  const removeCampaignItemHelper = (campaignId, itemId) => campaignActions.removeCampaignItem(
+    removeCampaignItemApi,
+    campaignId,
+    itemId
+  );
```
GuySartorelli, line 219: ~~Didn't exist before?~~
Ahh, was renamed from `removeCampaignItem`. Not clear why this renaming was done and the old name IMO was more appropriate - but it doesn't materially affect anything so leave as is.

Please just take care to comment on changes like this for future refactors.

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -387,114 +290,145 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
+      helpButtonRef.current.focus();
+    }
+
+    prevPropsRef.current = {
+      id: router.params.id,
+      view: router.params.view,
+      title,
+      showMessage,
+    };
+  }, [router.params.id, router.params.view, title, showMessage, setBreadcrumbs]);
```
GuySartorelli: Why is a function (`setBreadcrumbs`) in the dependencies array?

client/src/containers/CampaignAdmin/CampaignAdmin.js
```diff
@@ -387,114 +290,145 @@ By removing this item all linked items will be removed unless used elsewhere.`;
...
     const formBuilderProps = {
-      createFn: this.campaignListCreateFn.bind(this),
-      schemaUrl,
+      createFn: campaignListCreateFn,
+      schemaUrl: sectionConfig.form.EditForm.schemaUrl,
       identifier: 'Campaign.IndexView',
     };
 
-    const showSearch = hasFilters(this.state.filters) || this.state.showSearch;
+    const showSearchDisplay = hasFilters(filters) || showSearch;
```
GuySartorelli, line 372: Unclear why this was renamed and arguably not as good a name as before - but it doesn't materially change anything so okay to leave as is.

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -21,68 +21,33 @@ import CampaignAdminItem from './CampaignAdminItem';
...
+  BreadcrumbComponent = Breadcrumb,
+  DropdownItemComponent = DropdownItem,
+  FormActionComponent,
+  PreviewComponent,
+  ViewModeComponent,
+}) => {
+  const [stateError, setError] = useState(false);
+  const [errorCode, setErrorCode] = useState(0);
+
+  const loading = Object.keys(record).length !== 0;
```
GuySartorelli: Loading used to be state and ([according to line 553 of the original](https://github.com/silverstripe/silverstripe-campaign-admin/blob/bac309116c2d156d7432f29e9e4a324906b432ef/client/src/containers/CampaignAdmin/CampaignAdminList.js#L553)) in props - why not anymore?

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -21,68 +21,33 @@ import CampaignAdminItem from './CampaignAdminItem';
...
-          this.setState({
-            loading: false,
-            error: true,
-            errorCode: e.response.status
-          });
-        });
-    }
-  }
-
-  componentWillUnmount() {
```
GuySartorelli, line 75: So far in these refactors we've only had `componentDidMount`/`componentDidUpdate` so I need to know what the replacement is for `componentWillUnmount`.

I've found the answer but for any future new lifecycle functions we are replacing please save me the trouble by providing some links or even at the very least saying "this was replaced with a return statement in `useEffect`"

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -444,126 +379,145 @@ class CampaignAdminList extends Component {
...
         </div>
       </div>
     );
-  }
+  };
+
+  useEffect(() => {
+    setBreadcrumbs();
+    // Only load record if not already present
+    if (!loading) {
```
GuySartorelli: Depending on discussion in another comment thread about why this isn't in state anymore, this might not be an appropriate change - it used to be specifically `if (!this.isRecordLoaded()) {`

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -350,25 +287,25 @@ class CampaignAdminList extends Component {
 
     previewClasses = classNames(previewClasses);
 
-    if (loading) {
+    if (!loading) {
```
GuySartorelli: Why did this condition get flipped?

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -444,126 +379,145 @@ class CampaignAdminList extends Component {
...
+    };
+  }, []);
 
   /**
    * Renders a list of items in a Campaign.
    *
    * @return object
    */
-  render() {
-    const { campaign: { changeSetItemId }, campaignId, record: campaign } = this.props;
```
GuySartorelli, line 459: Note that the `record` prop was being renamed to `campaign` here - that means that any usage of `campaign` in this function probably needs to be updated to refer to `record` now or it'll be referring to the wrong prop.

client/src/containers/CampaignAdmin/CampaignAdminList.js
GuySartorelli: Why was `this.props.loading` changed to `!loading` (i.e. flipped condition)?

client/src/containers/CampaignAdmin/CampaignAdminList.js
```diff
@@ -444,126 +379,145 @@ class CampaignAdminList extends Component {
...
+
+  const loadingSpinner = !loading && [
+    <div key="overlay" className="cms-content-loading-overlay ui-widget-overlay-light" />,
+    <div key="spinner" className="cms-content-loading-spinner" />,
+  ];
+
+  return (
+    <div className={`fill-width campaign-admin__campaign ${selectedClass}`}>
+      {loadingSpinner}
+      {renderCampaignAdminListDetail(body)}
```
GuySartorelli, line 516: `itemLinks` argument was removed! Looks like there's no param to take it in the function being called so it's a correct change - please just remember to point out why changes like this are made.

GuySartorelli (changes requested):
Haven't tested locally yet as some of these changes may affect functionality.

client/src/components/IntroScreen/IntroScreen.js
emteknetnz: I don't think so, it's a DOM element so seems like it would be a weird dependency

client/src/containers/CampaignAdmin/CampaignAdminList.js
emteknetnz: It's the function that's returned in the useEffect() replacment - search for `Return cleanup function to reset new create flag on unmount`

https://robertmarshall.dev/blog/componentwillunmount-functional-components-react/ - explains it "How to manage componentWillUnmount with useEffect"

client/src/containers/CampaignAdmin/CampaignAdmin.js
emteknetnz: Fixed

client/src/containers/CampaignAdmin/CampaignAdmin.js
emteknetnz: Have refactored the useEffect hook that used this and got rid of the weird prevPropsRef

client/src/containers/CampaignAdmin/CampaignAdmin.js
emteknetnz: Has been refactored

client/src/containers/CampaignAdmin/CampaignAdmin.js
emteknetnz: Have changed these hooks, have removed the equivalent of the compnentDidMount hook since it's not required

client/src/containers/CampaignAdmin/CampaignAdminList.js
emteknetnz: I was getting endless errors when running `yarn test` related to some sort of "boundary" related to this. The original class component code is convoluted and really `const loading = Object.keys(record).length !== 0;` is all we need

emteknetnz: Opps just realised the condition is backwards because it used to be `isRecordLoaded`, now it's `loading` - have updated code accordingly

client/src/containers/CampaignAdmin/CampaignAdminList.js
emteknetnz: Updated

client/src/containers/CampaignAdmin/CampaignAdminList.js
emteknetnz: Good catch - there was one update to make `campaign.State === 'published'`

client/src/containers/CampaignAdmin/CampaignAdminList.js
emteknetnz: Updated

client/src/containers/CampaignAdmin/CampaignAdmin.js
GuySartorelli: > have removed the equivalent of the compnentDidMount hook since it's not required

It used to call `this.setBreadcrumbs(view, id, title);` if the `breadcrumbs` prop was empty. I'm probably missing some context but it doesn't look like that'll happen right now unless the component updates (which I think might not necessarily happen post mount)?

GuySartorelli (commented):
Oops, I forgot to hit "start review", Just one small thing below.
https://github.com/silverstripe/silverstripe-campaign-admin/pull/382#discussion_r2516649972

Works well locally.

client/src/containers/CampaignAdmin/CampaignAdmin.js
emteknetnz: It's not required, you can put a console.log() statement in there and you'll see it's called on mount

GuySartorelli (approved):
LGTM

